Issue #146, opened 2022-11-01
Would you accept a pull request for Face Recognition integration?


matiasdelellis:
First of all, thank you very much for the application. Love it. 😄 

On the other hand, I develop an unofficial facial recognition application for Nextcloud.
* https://github.com/matiasdelellis/facerecognition

It has much more development, but it will never be accepted as an official part since they prefer the automatic approach beyond the quality of the results, but I don't want to make a controversy about it either. I am writing here to see if you would be willing to accept a PR adding support for this application.

Now i'm doing tests, just modifying the backend queries to reuse the frontend.

![imagen](https://user-images.githubusercontent.com/733715/199301562-cc1c76ff-54dd-4d88-8a7c-99d646d14441.png)
![imagen](https://user-images.githubusercontent.com/733715/199365482-1bf862e7-bffe-4a1a-b1f9-0b3826bb0ae0.png)

However ideally i should add a new (and hidden by default) view, since the way it works is different (It is oriented to the faces, to avoid errors).
![imagen](https://user-images.githubusercontent.com/733715/199302127-fe114851-2aee-4ef0-8131-980a4cade549.png)

pulsejet:
Of course! In fact your app was my first choice for the integration, but feature parity with Photos had higher priority (which we almost have now).

There are just a few requirements before this can get merged:
1/ As you already noted, this needs to fit well into the TimelineQuery architecture. This should generally be pretty easy; I guess you already did most of this part.
2/ Basic feature parity with the recognize integration. So the face previews should work right (center the face), and things like renaming / merging (if you support this) face clusters from memories itself. Unfortunately, recognize uses a terrible webdav API to handle all this, so the two options we have are either having a similar webdav API for facerecognition, or duplicating some UI parts to call the existing API. I strongly prefer the latter since webdav sucks.
3/ We do need to support both backends. This is very easy -- if recognize is enabled with face recognition enabled, show it as People. If recognize isn't enabled or it's face feature is disabled, show faceregnition as People if enabled. If both apps are enabled, just have two tabs, `People (recognize)` and `People (faces)` (subtitle if possible). People might want to run both apps and compare the results (I do).

matiasdelellis:
Hi @pulsejet 
Thank you very much for allowing me to work on it... 😄 

We agree on everything... Nextcloud took the unix saying, "everything is a file" to the extreme "everything is a dav entry". 🙈 😅 

[...]

PR #148, opened 2022-11-02
Initial implementation of the Face Recognition app integration

matiasdelellis:
At this point, we completely reuse the Recognize view, and for now, Recognize has priority. If it is installed, it shows the results of this application.

So if recognize is disabled and Face Recognition is enabled, you can see the tagged people on that app, you can see the photos of these people, and the rectangles of the faces.

Still can't add names, hide people, or merge clusters with the current view since it is intended for the recognize DAV api.

We probably need to add a custom view for this app. Meanwhile, the side panel that works in both Files and Photos also works in Memories. You can add the names there, hide the people, and it should probably only be done there.

### DIFF
--- a/lib/Db/TimelineQuery.php
+++ b/lib/Db/TimelineQuery.php
@@ -11,8 +11,8 @@ class TimelineQuery
 {
     use TimelineQueryAlbums;
     use TimelineQueryDays;
-    use TimelineQueryFaces;
-    use TimelineQueryPersons;
+    use TimelineQueryFaceRecognition;
+    use TimelineQueryRecognizePeople;
     use TimelineQueryFilters;
     use TimelineQueryFolders;
     use TimelineQueryLivePhoto;
--- a/lib/Db/TimelineQuery.php
+++ b/lib/Db/TimelineQuery.php
@@ -11,11 +11,11 @@ class TimelineQuery
 {
     use TimelineQueryAlbums;
     use TimelineQueryDays;
-    use TimelineQueryFaceRecognition;
     use TimelineQueryFilters;
     use TimelineQueryFolders;
     use TimelineQueryLivePhoto;
-    use TimelineQueryRecognizePeople;
+    use TimelineQueryPeopleFaceRecognition;
+    use TimelineQueryPeopleRecognize;
     use TimelineQueryTags;
 
     protected IDBConnection $connection;
--- a/lib/Db/TimelineQuery.php
+++ b/lib/Db/TimelineQuery.php
@@ -12,6 +12,7 @@ class TimelineQuery
     use TimelineQueryAlbums;
     use TimelineQueryDays;
     use TimelineQueryFaces;
+    use TimelineQueryPersons;
     use TimelineQueryFilters;
     use TimelineQueryFolders;
     use TimelineQueryLivePhoto;
--- a/lib/Db/TimelineQuery.php
+++ b/lib/Db/TimelineQuery.php
@@ -12,10 +12,10 @@ class TimelineQuery
     use TimelineQueryAlbums;
     use TimelineQueryDays;
     use TimelineQueryFaceRecognition;
-    use TimelineQueryRecognizePeople;
     use TimelineQueryFilters;
     use TimelineQueryFolders;
     use TimelineQueryLivePhoto;
+    use TimelineQueryRecognizePeople;
     use TimelineQueryTags;
 
     protected IDBConnection $connection;
--- a/lib/Db/TimelineQueryFaceRecognition.php
+++ b/lib/Db/TimelineQueryFaceRecognition.php
@@ -24,12 +24,18 @@ trait TimelineQueryFaceRecognition
         $personName = $personNames[1];
 
         // Join with images
-        $query->innerJoin('m', 'facerecog_images', 'fri',
+        $query->innerJoin(
+            'm',
+            'facerecog_images',
+            'fri',
             $query->expr()->eq('fri.file', 'm.fileid')
         );
 
         // Join with faces
-        $query->innerJoin('fri', 'facerecog_faces', 'frf',
+        $query->innerJoin(
+            'fri',
+            'facerecog_faces',
+            'frf',
             $query->expr()->eq('frf.image', 'fri.id')
         );
 
@@ -101,7 +107,7 @@ trait TimelineQueryFaceRecognition
 
         // Post process
         foreach ($faces as &$row) {
-            $row['id'] = $row['name'] ? $row['name'] : (int) $row['id'];
+            $row['id'] = $row['name'] ?: (int) $row['id'];
             $row['count'] = (int) $row['count'];
         }
 
@@ -132,13 +138,12 @@ trait TimelineQueryFaceRecognition
         // WHERE these photos are memories indexed
         $query->innerJoin('fri', 'memories', 'm', $query->expr()->eq('m.fileid', 'fri.file'));
 
+        $query->innerJoin('frf', 'facerecog_persons', 'frp', $query->expr()->eq('frp.id', 'frf.person'));
         if (is_numeric($previewId)) {
             // WHERE faces are from id persons (a cluster).
-            $query->innerJoin('frf', 'facerecog_persons', 'frp', $query->expr()->eq('frp.id', 'frf.person'));
             $query->where($query->expr()->eq('frp.id', $query->createNamedParameter($previewId)));
         } else {
             // WHERE faces are from name on persons.
-            $query->innerJoin('frf', 'facerecog_persons', 'frp', $query->expr()->eq('frp.id', 'frf.person'));
             $query->where($query->expr()->eq('frp.name', $query->createNamedParameter($previewId)));
         }
 
--- a/lib/Db/TimelineQueryFaceRecognition.php
+++ b/lib/Db/TimelineQueryFaceRecognition.php
@@ -8,11 +8,11 @@ use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\Folder;
 use OCP\IDBConnection;
 
-trait TimelineQueryPersons
+trait TimelineQueryFaceRecognition
 {
     protected IDBConnection $connection;
 
-    public function transformPersonFilter(IQueryBuilder &$query, string $userId, string $personStr)
+    public function transformFaceRecognitionFilter(IQueryBuilder &$query, string $userId, string $personStr)
     {
         // Get title and uid of face user
         $personNames = explode('/', $personStr);
@@ -42,7 +42,7 @@ trait TimelineQueryPersons
         ));
     }
 
-    public function transformPersonRect(IQueryBuilder &$query, string $userId)
+    public function transformFaceRecognitionRect(IQueryBuilder &$query, string $userId)
     {
         // Include detection params in response
         $query->addSelect(
@@ -55,7 +55,7 @@ trait TimelineQueryPersons
         );
     }
 
-    public function getPersons(Folder $folder, bool $show_clusters = false, bool $show_hidden = false)
+    public function getFaceRecognitionPeople(Folder $folder, bool $show_clusters = false, bool $show_hidden = false)
     {
         $query = $this->connection->getQueryBuilder();
 
@@ -108,7 +108,7 @@ trait TimelineQueryPersons
         return $faces;
     }
 
-    public function getPersonPreviewDetection(Folder &$folder, $previewId)
+    public function getFaceRecognitionPreview(Folder &$folder, $previewId)
     {
         $query = $this->connection->getQueryBuilder();
 
@@ -201,7 +201,7 @@ trait TimelineQueryPersons
     }
 
     /** Convert face fields to object */
-    private function processPerson(&$row, $days = false)
+    private function processFaceRecognitionDetection(&$row, $days = false)
     {
         if (!isset($row) || !isset($row['image_width'])) {
             return;
--- a/lib/Db/TimelineQueryFaces.php
+++ b/lib/Db/TimelineQueryFaces.php
@@ -14,7 +14,7 @@ trait TimelineQueryFaces
 
     public function transformFaceFilter(IQueryBuilder &$query, string $userId, string $faceStr)
     {
-        // Get title and uid of face user
+        // Get name and uid of face user
         $faceNames = explode('/', $faceStr);
         if (2 !== \count($faceNames)) {
             throw new \Exception('Invalid face query');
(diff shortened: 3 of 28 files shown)


lib/Db/TimelineQueryFaces.php
```diff
@@ -14,7 +14,7 @@ trait TimelineQueryFaces
 
     public function transformFaceFilter(IQueryBuilder &$query, string $userId, string $faceStr)
     {
-        // Get title and uid of face user
```
matiasdelellis, line 17: D'Oh!. Unnecessary change.

matiasdelellis:
Backend ready to request comments...  I guess it's a good proof of concept. 😅

imagen
imagen

It shows the people that I have already named, and the clusters that I have not named yet. When clicking on these people or clusters, it shows the photos correctly, but still no change is implemented in the fronted.

According to your comments in #146, I have to make a new view. Since it is working in the current view, in principle it is easy, but now i have doubts with the names to use for the components.

For example, I have to duplicate all the modals to adapt them to our API. I'm not very savvy with VUE. Duplicating code would be easy, but if there is a way to reuse it, I'd appreciate your help. 🤔

lib/Db/TimelineQuery.php
```diff
@@ -11,7 +11,8 @@ class TimelineQuery
 {
     use TimelineQueryAlbums;
     use TimelineQueryDays;
-    use TimelineQueryFaces;
+    use TimelineQueryFaceRecognition;
+    use TimelineQueryRecognizePeople;
```
pulsejet, line 15: Maybe `TimelineQueryPeopleFaceRecognition` and `TimelineQueryPeopleRecognize` for semantic grouping. I really shouldn't have done all of this trait stuff...

lib/Db/TimelineQueryFaceRecognition.php
```diff
@@ -0,0 +1,222 @@
...
+        if (2 !== \count($personNames)) {
+            throw new \Exception('Invalid person query');
+        }
+
+        $personUid = $personNames[0];
+        $personName = $personNames[1];
+
+        // Join with images
+        $query->innerJoin('m', 'facerecog_images', 'fri',
+            $query->expr()->eq('fri.file', 'm.fileid')
```
pulsejet, line 28: I'm yet to test this, but I asssume `fri.file` is `BIGINT(20)`? Otherwise this is gonna be horribly slow.

lib/Db/TimelineQueryFaceRecognition.php
```diff
@@ -0,0 +1,222 @@
...
+        );
+
+        // Join with faces
+        $query->innerJoin('fri', 'facerecog_faces', 'frf',
+            $query->expr()->eq('frf.image', 'fri.id')
```
pulsejet, line 33: Similar question - assuming these have the same type?

lib/Db/TimelineQueryFaceRecognition.php
```diff
@@ -0,0 +1,222 @@
...
+        $cursor = $this->executeQueryWithCTEs($query);
+        $previews = $cursor->fetchAll();
+        if (empty($previews)) {
+            return null;
+        }
+
+        // Score the face detections
+        foreach ($previews as &$p) {
+            // Get actual pixel size of face
+            $iw = min((int) ($p['image_width'] ?: 512), 2048);
```
pulsejet, line 165: Any way this code could be de-duplicated? I assume the logic is unchanged/similar?

src/components/frame/Tag.vue
```diff
@@ -85,7 +85,7 @@ export default class Tag extends Mixins(GlobalMixin) {
   getPreviewUrl(photo: IPhoto) {
     if (this.isFace) {
       return generateUrl(
-        "/apps/memories/api/faces/preview/" + this.data.fileid
+        "/apps/memories/api/recognize/people/preview/" + this.data.fileid
```
pulsejet, line 88: You do need to call the `facerecognition` API here. I assume that's something WIP?

pulsejet (commented):
1. I think the backend looks solid - just some minor comments.
2. Assuming you're still working the Vue part? Don't need everything for this PR; just showing the faces list and contents is good to go.
3. Also assuming face recognition already has the relavant indices for the queries being made here. These tend to get super-expensive; better test with `EXPLAIN` to be sure. As a rule of thumb any queries for < 5k photos must finish within a few ms at most. But I could also look into this later.

Btw, I'm travelling internationally right now so might be slow to respond.

pulsejet:
For the lint fail - you can just run `make php-lint` when you're done to fix everything.

pulsejet:
> For example, I have to duplicate all the modals to adapt them to our API. I'm not very savvy with VUE. Duplicating code would be easy, but if there is a way to reuse it, I'd appreciate your help. 🤔

Missed this comment. I'm more inclined towards `switch case` inside the modals to make the right calls. Are there situations where this isn't possible?

lib/Db/TimelineQueryFaceRecognition.php
matiasdelellis: Ohhh.. Good catch..
I never realized it. I will add a migration in my application to adapt it. 😉

lib/Db/TimelineQueryFaceRecognition.php
matiasdelellis: Here both types are equal... 
Though ideally maybe I should change them to bigint as well... Maybe later. It is not necessary here.

lib/Db/TimelineQueryFaceRecognition.php
matiasdelellis: Face Recognition saves the detections as top, bottom, right, left. I have intentions to change to width, height, but it will never be compatible with recognize, which uses these parameters as percentages of the image.

This is probably one of the biggest recognize errors, since the detections must be pixel perfect when getting the face alignments and descriptors.
However, I understand that this decision is due to the fact that it is what face-api offers.

So, I will always have to calculate the percentages, but (except the confidence, which is a score in itself) the subsequent calculation of the score could be reused.

src/components/frame/Tag.vue
matiasdelellis: > You do need to call the facerecognition API here. I assume that's something WIP?

You say Faceregnition own api? Just changed it, to be consistent and then use an "/apps/memories/api/facerecognition/people/preview/" endpoint..

If we are going to reuse and not duplicate components as you mentioned, here i will add a switch case to select the correct endpoint.

P.S: The api entry "/apps/memories/api/facerecognition/people/preview/" already works, and my tests are changing the method in the routes.php file, and other small changes. I'll comment on it later. 🤔

lib/Db/TimelineQueryFaceRecognition.php
pulsejet: Converting them either way is fine -- if it makes more sense, you can convert recognize's percent values to absolute ones and call the common function.

matiasdelellis:
A little progress...
imagen
imagen

It allows to visualize everything correctly but still no works any action in facerecognition.

matiasdelellis:
# TODO:

Rename person and Merge with different Person:
- [ ] **PUT:** /ocs/v2.php/apps/facerecognition/api/2.0/cluster/317878?name=Name
- [ ] **PUT:** /ocs/v2.php/apps/facerecognition/api/2.0/person/Penny?name=Not+Penny

Remove person:
- [ ] **PUT:** /ocs/v2.php/apps/facerecognition/api/2.0/cluster/317878?visible=false
- [ ] **PUT:** /ocs/v2.php/apps/facerecognition/api/2.0/person/Penny?visible=false

The API is easy... I'm looking at how to do it elegantly,

Others todos..
- [ ] This will depend on a Face Recognition update. https://github.com/matiasdelellis/facerecognition/pull/608 🤔
- [ ] Maybe we must check Nextcloud Capabilities to know if the user activated the application. 🤔 
- [ ] Make it show only the current model. 🤦🏻 It should be easy.. 😅 
- [ ] Sort by number of photos. 😅 
- [ ] Rebase on master..

lib/Db/TimelineQueryFaceRecognition.php
matiasdelellis: Done on https://github.com/matiasdelellis/facerecognition/pull/608/commits/bfdaab2cd24c58b737c65351e02493fc33055d54

lib/Db/TimelineQueryFaceRecognition.php
matiasdelellis: Partially done..
https://github.com/matiasdelellis/facerecognition/pull/608/commits/bfdaab2cd24c58b737c65351e02493fc33055d54

Now I have to adapt the queries.

src/components/Timeline.vue
```diff
@@ -668,6 +687,27 @@ export default class Timeline extends Mixins(GlobalMixin, UserConfig) {
...
   }
 
+  /* Get a friendly description of empty view */
+  get emptyViewDescription() {
+    switch (this.$route.name) {
+      case "facerecognition":
+        if (this.config_facerecognitionEnabled)
+          return this.t('memories', 'You will find your friends soon. Please, be pacient.')
+        else
+          return this.t('memories', 'Face Recognition is disabled. Enabled in settings to find your friends.')
```
pulsejet, line 697: "Enabled" --> "Enable"

src/components/Timeline.vue
```diff
@@ -668,6 +687,27 @@ export default class Timeline extends Mixins(GlobalMixin, UserConfig) {
     return head.name;
   }
 
+  /* Get a friendly description of empty view */
+  get emptyViewDescription() {
+    switch (this.$route.name) {
+      case "facerecognition":
+        if (this.config_facerecognitionEnabled)
+          return this.t('memories', 'You will find your friends soon. Please, be pacient.')
```
pulsejet, line 695: "pacient" --> "patient"

pulsejet (commented):
LGTM overall, just a couple of questions:
1. The face top matter has some recognize-specific controls (e.g. delete face). Don't we need to hide these until functional for face recognition?
2. We need at least the rename function, otherwise I'm afraid new users might be scared off using face recognition due to missing UX.

src/components/Timeline.vue
matiasdelellis: D'OH!. Done..

src/components/Timeline.vue
matiasdelellis: D'OH!. Done..

matiasdelellis:
Hi @pulsejet 

> * The face top matter has some recognize-specific controls (e.g. delete face). Don't we need to hide these until functional for face recognition?

The only thing missing is the merge dialog that would be easy, but it is a "big" rewrite of  `FaceMergeModal.vue`... we just need to add something like `await dav.renamePeopleFaceRecognition(this.oldName, this.name);` but the dialog originally does a lot of checks that are unnecessary here.. 🤔  You can see it?

On the other hand, I just discovered "Move to another person" and "Remove from person" that are hidden correctly. The app is also easy..
* PUT: /index.php/apps/facerecognition/api/2.0/face/{faceId}/detach?name=New+Name (Move to another person)
* PUT: /index.php/apps/facerecognition/api/2.0/face/{faceId}/detach (Remove from person)

But we need the id of the face.. Not the file... 😥 

I have to think about it, but in principle it is another task that really works from the side bar..

pulsejet:
@matiasdelellis sorry I missed your previous comment.

> The only thing missing is the merge dialog that would be easy, but it is a "big" rewrite of `FaceMergeModal.vue`... we just need to add something like `await dav.renamePeopleFaceRecognition(this.oldName, this.name);` but the dialog originally does a lot of checks that are unnecessary here.. 🤔 You can see it?

Maybe just create two functions `clickFaceRecognize` and `clickFaceFR` that do their stuff? Then have `clickFace` appropriately dispatch.

> **EDIT:** The truth is that I would prefer to add autocompletion to the name dialog. I find it much more useful than the merge dialog.. 😅
> 
> On the other hand, I just discovered "Move to another person" and "Remove from person" that are hidden correctly. The Api is also easy..
> 
> * PUT: /index.php/apps/facerecognition/api/2.0/face/{faceId}/detach?name=New+Name (Move to another person)
> * PUT: /index.php/apps/facerecognition/api/2.0/face/{faceId}/detach (Remove from person)
> 
> But we need the id of the face.. Not the file... 😥

Is the fileid in the PUT body? Again in that case you just need a separate function, then dispatch to the right one. You can get the face ID from the route (if not, an ugly hack is to add it to the route to begin with, say `<id>-<name>` instead of just `<name>` in the route.

`SelectionManager` is getting a quite bloated tbh, but let's worry about that later.

> I have to think about it, but in principle it is another task that really works from the side bar..

The sidebar only works for individual images, which isn't the easiest to use in case the AI screws up (we have to assume it can happen) ...

BTW, marking this for the next version.

pulsejet:
@matiasdelellis don't worry about the merge conflicts. I'll take care of that.

pulsejet:
Manually merged with this https://github.com/pulsejet/memories/commit/aeffe628f2fc6d0c99bd3e4537497eaf96d8fbc0

pulsejet:
@matiasdelellis some general comments from my testing of facerecognition

1. The results are good. I like the parameter tuning; feels like a lot more control. I did have to slightly tweak them to get the best results.
2. pdlib is a hassle to set up. It didn't even compile out of the box with dlib:master, had to do some minor tweaking in the Makefile (it uses `std=c++11` while dlib needs at least 14; setting CXXFLAGS seems to have no effect).
3. I'd worry a bit about pdlib's state of maintenance in general. The last commit was 2020.
4. It might be useful to provide a (say partial) Dockerfile that sets up everything in a container.
5. I noticed facerecognition uses only one core, unless I missed some configuration. That's kinda unfortunate since I have 64 and yet I've to wait 5 days (?) for it to finish.

matiasdelellis:
Woow... Thanks you very much!!. 😃 🎉 

>    The results are good. I like the parameter tuning; feels like a lot more control. I did have to slightly tweak them to get the best results.

I'm glad you see it that way. There are some who get scared by these configurations..

>    pdlib is a hassle to set up. It didn't even compile out of the box with dlib:master, had to do some minor tweaking in the Makefile (it uses std=c++11 while dlib needs at least 14; setting CXXFLAGS seems to have no effect).
>    I'd worry a bit about pdlib's state of maintenance in general. The last commit was 2020.

It's disappointing, and I already forked once until the author accepted the necessary changes... 😞 
I should probably do it again, and maybe expand it with other things.. 🤔 😉 

>    It might be useful to provide a (say partial) Dockerfile that sets up everything in a container.

It is impossible to meet all users, so I promote the documentation.

* https://github.com/matiasdelellis/facerecognition/wiki/Docker

>    I noticed facerecognition uses only one core, unless I missed some configuration. That's kinda unfortunate since I have 64 and yet I've to wait 5 days (?) for it to finish.

Personally I prefer a core at 100% for a longer time, than the whole microprocessor at 100% for a few hours. 🙈 😅 
On my personal server, I only have 4 cores and 8 GB of ram, and maybe that's why I never had enough interest in doing it, likewise, it is one of the pending tasks for a version 1.0..

Issue #99, opened 2022-10-23
FATAL: Encryption is enabled. Aborted.


TonyNoIT:
After I tried to - _sudo -u www-data php ./occ memories:index_
Give error 
FATAL: Encryption is enabled. Aborted.

Is this module works with standard encryption on Nextcloud?

pulsejet:
> Is this module works with standard encryption on Nextcloud?

No. Because memories does server-side processing, it cannot support end-to-end encrypted files without leaking information.

eltos:
@pulsejet could you elaborate a bit more on this?  
Nextcloud features two kinds of encryption: 
- Server-side encryption: files on the drive are encrypted, but the server has the keys when a user is logged in
- End-to-end encryption: No way for the server to decrypt files

Is it only the latter or both that are not supported?  
Is it just the occ command for initial indexing of existing files, or the memories app in general not supporting it?  
Is there an alternative way for initial indexing (e.g. deleting and re-uploading all files)?

eltos:
Just tested it with _server-side_ encryption: the occ command fails, but all images uploaded _after_ installation of the Memories app _do_ show up in the timeline. So maybe all it needs to fully support _server-side_ encryption is a method to do the indexing on a per-user basis, e.g. from the welcome screen when a user first opens the app?

pulsejet:
I assumed the issue is about e2e encryption (maybe incorrectly). Not sure how the internals of server-side encryption work, but I'm willing to accept a pull request if we can support this.

eltos:
Might be worth taking a look how the [metadata app](https://github.com/gino0631/nextcloud-metadata) does it. Maybe there is even a way to interface with it directly.

TonyNoIT:
> 

That what I did for my files…
If files uploaded after - everything work 🤷‍♂️

eltos:
Well, actually even the `occ memories:index` command works perfectly fine with server-side encryption. Just comment out that return statement in [lib/Command/Index.php](https://github.com/pulsejet/memories/blob/86de94c210914b5074c0cb073d9164ca900c7f4f/lib/Command/Index.php#L194-L198)
```php
if ($this->encryptionManager->isEnabled()) {
    error_log('FATAL: Encryption is enabled. Aborted.');
    return 1;
}
```
[...]

PR #175, opened 2022-11-07
Allow server-side encryption

eltos:
Closes https://github.com/pulsejet/memories/issues/99


pulsejet:
Thanks for digging into this. This looks fine to me, as long as you tested it. If `updateExifDate` also works as you mentioned, you can just move the same check to `Utils` and call it from `ImageController` (and get rid of the check in `updateExifDate`).

pulsejet:
Just to be sure, this is tested on at least Nextcloud 24?

eltos:
Tested it on NC 24 back then, just tested on NC 25.0.1 again.

pulsejet:
Thank you! I'll take it from here.

PR #221, opened 2022-11-19
Report progress during indexing

eltos:
Besides adding a percentage progress, this also exploits the ConsoleSectionOutput capability of Symfony to simplify the carriage return gymnastics and to fix it for paths exceeding the console window width.

Tested on NC 25.0.1


eltos:
That's odd, a count is always an int. The only way this can become a float simply my multiplication with another int, is if the number exceeds the integer capacity. Which is 9223372036854775807 on that 64 bit system. The guy must have some files 🤣

pulsejet:
I don't know what happened either 🤷🏻

PR #245, opened 2022-11-23
Add explicit .PHONY rules

Sigmanificient:
Mark all rules that doesn't depend on files as PHONY to avoid any future conflict.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,8 @@ npm-init:
 npm-update:
 	npm update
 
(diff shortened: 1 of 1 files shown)


pulsejet:
Hopefully I'll see more ;)

Sigmanificient:
> Hopefully I'll see more ;)

I love vue, but i'm not got enough with typescript unfortunately

pulsejet:
Give it a go regardless of this project. After I used typescript for a while, I'm never looking back. Plain JS seems like playing with fire now.

PR #260, opened 2022-11-26
Timeline for each folder

keyeh:
Thank you for creating this app. The recycler scrolling and video transcoding is absolutely brilliant!

I've been using digiKam and PiGallery2 for many years and they both support viewing a folder's contents recursively. It's a super useful feature for people that organize by folder structure. If you have many photos from different sources it's more manageable to have smaller timelines per subfolder.

In [digiKam it's a global setting](https://user-images.githubusercontent.com/5600824/204072098-2e5a8344-69e7-4e4a-b546-1d45ed7f58af.png) and in [PiGallery2 it's a button](https://user-images.githubusercontent.com/5600824/204072205-9349d756-cd15-4cef-b3c3-249cec829ee3.png) to recursively see all photos in a selected folder. This functionality almost exists in memories already, it just needs to be connected to the frontend.

This PR adds adds a button in `FolderTopMatter` so when the user is viewing a folder they can choose to see Timeline of the folder's contents. When the button is clicked it toggles the URL query param `?recursive=1` which is then passed to the API.

See screen recording below.

Unfortunately I am not a Vue developer, so please help if my code is not organized correctly 😅 

[screen-capture (3).webm](https://user-images.githubusercontent.com/5600824/204072531-97b768fe-78d5-41db-b270-907d88553bc4.webm)


src/components/Timeline.vue
```diff
@@ -240,7 +240,7 @@ export default class Timeline extends Mixins(GlobalMixin, UserConfig) {
 
   @Watch("$route")
   async routeChange(to: any, from?: any) {
-    if (from?.path !== to.path) {
+    if (from?.path !== to.path || from.query.recursive !== to.query.recursive) {
```
pulsejet, line 243: Maybe for the query in general rather than recursive specifically?

pulsejet (commented):
Very nice, thanks for the contribution!

I've a couple of questions / comments:

1. Would it not make sense to also show the folders at the top in the timeline view? It seems a bit strange to switch to folders view and only then be able to go to the nested folder, no?
2. I unfortunately was lazy and assumed at many places that `recursive` is equivalent to `timeline`. There are a few things we don't want to allow the user to do anywhere except the timeline (archiving photos, for instance). I do believe _most_ of these things got fixed with v4.8.0 but the implications of setting `isRecursive` to true need to be checked very carefully.

src/components/Timeline.vue
keyeh: Sure, I can change it to `from.query !== to.query` which also works, or do you have a suggestion for another way?

keyeh:
Hi, thanks for looking at this. Yes it would be better if there's a way to navigate folders while in the timeline. Otherwise the user has to switch back to folder view to navigate, but I don't think it's strange since PiGallery2 behaves like that too.

I'll use this for a while and try to figure out folder navigation. I would prefer having a folder tree or dropdown instead of putting the folders above images in timeline view though. but that's more work.

src/components/Timeline.vue
pulsejet: That sounds good to me, so long as it works.

pulsejet:
I'm fine either way, since I never used folders this way. Just trying to figure out what makes best sense.

PR #321, opened 2022-12-31
Add move to folder feature

### LINKED ISSUE #204: [Feature Request] Able to copy / move from timeline or folders
The timeline nicely connects newly added pictures which join the fray thanks to auto-uploading or other means. Every so often, I'd like to move certain pictures to other folders, for example if a picture of my kid is automatically uploaded, I'd like to move it to the folder where I keep all pictures of this kind.

I can go the manual way and open the folder containing the picture through the context menu and do the move/copy there, but doing it from the timeline and/or folders would streamline this.

ahaltindis:
I've been looking for a 'perfect' self hosted gallery app for some time, and this is the best I could find so far. So thanks for this project!

While I am using this app though, I notice couple of features fell short, which are important for my use-case. So moving multiple photos from timeline and folders to another folder is the one I need at the moment (effectively fixes #204 - only move part though). So I decided to implement this. I understand that if this won't be prioritized (I'll use my fork anyway 😄) or cancelled all together.

Couple of notes:
- Webdav natively supports `MOVE`, so just using it as similar to Nextcloud files.
- If the target exists, it'll fail. (again, same behavior with Nextcloud files).
- Not enabling move in albums and archive as it would probably be doing something that user didn't intend.
- Not adding move operation to context menu of photo view as that doesn't seem very useful (at least to me) so didn't want to add unnecessary noise.
- Using very similar approaches to adding to album and deleting photos.

### DIFF
--- a/src/components/SelectionManager.vue
+++ b/src/components/SelectionManager.vue
@@ -44,6 +44,7 @@
       :updateLoading="updateLoading"
     />
     <AddToAlbumModal ref="addToAlbumModal" @added="clearSelection" />
+    <MoveToFolderModal ref="moveToFolderModal" @moved="refresh" />
   </div>
 </template>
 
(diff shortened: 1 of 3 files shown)


Gymnae:
Thanks for creating a PR for this, just today I was hoping this feature would come :)

pulsejet:
_Very_ nice (one of the rare times I've had a PR where I changed _nothing_😄).

Of course this will break when the user has e.g. a shared folder or something as their timeline root. But that breaks a lot of things anyway, so wouldn't bother with it for now. Ideally we need the permissions flags for each file so we can make checks on the UI side, but not sure how feasible this is, peformance-wise.

PR #332, opened 2023-01-07
Implement prepared statements in TimelineWrite

pHr34kY:
I noticed that when running `occ memories:index`, TimelineWrite would build a new query for every CRUD operation on every file processed. Creating an `IPreparedStatement` in the constructor builds the query once and re-uses it for each file processed. When re-scanning an image directory, this improved the overall speed by around 20%.

Before:
```
==========================================
Checked 289903 files of 4 users in 50.25 sec
6217 not valid media items
4 .nomedia folders ignored
283682 skipped because unmodified
0 (re-)processed
==========================================
```

After:
```
==========================================
Checked 289903 files of 4 users in 41.062 sec
6217 not valid media items
4 .nomedia folders ignored
283682 skipped because unmodified
0 (re-)processed
==========================================
```

### DIFF
--- a/lib/Db/TimelineWrite.php
+++ b/lib/Db/TimelineWrite.php
@@ -6,8 +6,8 @@ namespace OCA\Memories\Db;
 
 use OCA\Memories\AppInfo\Application;
 use OCA\Memories\Exif;
-use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\DB\IPreparedStatement;
+use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\File;
 use OCP\IDBConnection;
 use OCP\IPreview;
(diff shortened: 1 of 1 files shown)


pHr34kY:
I should note, I've tested all operations on MySQL. Nothing should cause issues in other db engines, but it would definitely be worth checking.

pulsejet:
Nice. I had no idea the query builder is so slow.

pHr34kY:
The query builder is quick enough, but a newly built query needs to have an execution plan mapped out by the SQL backend. The overhead only impacts the first execution.

Also, I see the linter hates me. Do I need to try and clean that up, or do you typically do this?

pulsejet:
Hi again. First, thanks! However, I'm not fully convinced the added complexity is worth the performance improvement.
1. By preparing the statements in the constructor, they will be prepared for every request. Most of these won't be used for every request, so that's wasted CPU cycles.
2. This will benefit only the index command, which ideally should be called only once (ever) except in special cases. As such it seems premature to optimize index performance.
3. Regardless, the performance when indexing individual files will be the same / slower due to point 1. Because these are indexed immediately on upload, a new instance of TimelineWrite is created for each file.

pHr34kY:
I ran both versions over my ~300,000 image collection and found that the difference was only a few seconds over an hour of processing. It seems that 99.9% of time spent indexing new files is lost within Nextcloud's core code.

PR #363, opened 2023-01-21
Fix an issue where shared videos cannot be loaded

### LINKED ISSUE #361: Unable to play videos shared by another user
First of all, thank you for this wonderful plugin. My wife and I have been using it for the past few months and we really love it!

I recently noticed an issue where videos shared by another user cannot be played in the Memories view with the error message: `The media could not be loaded, either because the server or network failed or because the format is not supported`. I can easily watch my own videos and my wife hers, but we cannot play each other's videos. It's not a codec issue because we have the same phones. When we find the same file in folder the videos play normally. 

This happens consistently on the latest version of Nextcloud v25.0.3 and Memories v4.10.1. Video transcoding is not enabled.

I'm not sure if this bug was introduced in a recent update to v4.10, but it's possible.

jeremija:
Closes #361.

Not sure if this has any other implications, but it fixes the described issue on my instance.

### DIFF
--- a/lib/Controller/DownloadController.php
+++ b/lib/Controller/DownloadController.php
@@ -126,13 +126,17 @@ class DownloadController extends ApiBase
             return new JSONResponse([], Http::STATUS_NOT_FOUND);
         }
 
+        // Get the UID of the logged in user so that we can use the correct
+        // root path for files shared by another user.
+        $uid = $this->getUID();
+
         // Get DAV location of file
-        $userFolder = $this->rootFolder->getUserFolder($file->getOwner()->getUID());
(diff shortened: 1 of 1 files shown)


pulsejet:
Fixed with https://github.com/pulsejet/memories/commit/b9722e925c4cdb026e259a50f73c8bb4d390cc7e

We can't do this since it'll break public shares (uid blank). The issue was that the user context of the file returned by getUserFile might be either the current user or the owner. So we can't call `getRelativePath` on `getPath` from a different user's context path,

jeremija:
Thanks for explaining and for a quick fix! I'll try to test it shortly!

Issue #121, opened 2022-10-26
Display map of photos


meichthys:
Potential Solution: The nextcloud maps app displays the locations of photos, so we could just link to that.

On a side note, the metadata nextcloud app does display the geolocation of photos. The address is clickable which opens a small dialog with an openstreetmap view of the location the photo was taken.

pulsejet:
Sounds good. Not sure what it'll take to get the sidebar to work right though.

jacotec:
@pulsejet Sorry to jump into an already closed issue, but the NC Maps application is unfortunately horrible maintained and the performance is absolutely worse.

It currently does not work at all and even nearly crashes the server in case a user has more than a few hundred photos and nobody seems to care about: https://github.com/nextcloud/maps/issues/865
So I (and some others) have been forced to disable it to ensure a stable server operation.

Your app is an absolute boost in performance compared to Photos and if you see any chance to integrate a maps feature directly into Memories with your performance knowledge, it would be absolutely great!

meichthys:
That being said, you may be interested in the screenshot here that seems to show metadata (and a map)!
https://github.com/pulsejet/memories/issues/68#issuecomment-1306440766

jacotec:
> @jacotec I'm using the maps app with about 10k photos without much trouble. It does take about 10 seconds to load the thumbnails, but it doesn't seem to be that bad. 

Maybe it works somehow with 10k photos, but it definitely does not work at all with >50k. And it's not running on a Pi, it's a Dell R720 and NC has 10 vCPU and 16GB RAM assigned.

jacotec:
> That being said, you may be interested in the screenshot here that seems to show metadata (and a map)! [#68 (comment)](https://github.com/pulsejet/memories/issues/68#issuecomment-1306440766)

Yap, I saw this ... but the map is the location of a single photo (although I extremely love it!). What we need is a real map with all the photos and a good reaction ... similar to what I have on my iPad in Apple Photos.

I was using another open source project for photos before, Picapport. Very responsive map display for >100k photos. But as we all have in the Nexcloud, it's useless to sync a second thing just for map display.

jacotec:
> Of course there's no point reimplementing all the features from maps (that's _definitely_ out of scope), so we need to first make a list of which features are specifically the most useful and relevant.

[...]

PR #376, opened 2023-01-26
Show photos taken in the visible area of the map

RaymondHuang210129:
First of all, Thanks for your great project! This app allows me to migrate my photos from Google to my self-hosted NAS without compromising the user experience.

I like the location feature in Google Photos because I can locate my travel photos by dragging the map to specific places.
Thus I would like to contribute some codes to make this feature available in this app.

I added another page called 'Locations' which shows the map on the top, and the timeline shows the photos taken within the visible range in the map.

I guess the only way to change schema automatically is to have the app updated and run the migration script, so I manually added two columns (latitude, longitude) to the table `oc_memories` and runs the below query to fill each row:

```sql
ALTER TABLE oc_memories ADD latitude double DEFAULT NULL;
ALTER TABLE oc_memories ADD longitude double DEFAULT NULL;
UPDATE IGNORE oc_memories SET longitude = IF(exif REGEXP '(?<=GPSLatitude\":).*?(?=,)' AND exif REGEXP '(?<=GPSLongitude\":).*?(?=,)', REGEXP_SUBSTR(exif, '(?<=GPSLongitude\":).*?(?=,)') + 0.0, NULL);
UPDATE IGNORE oc_memories SET latitude = IF(exif REGEXP '(?<=GPSLatitude\":).*?(?=,)' AND exif REGEXP '(?<=GPSLongitude\":).*?(?=,)', REGEXP_SUBSTR(exif, '(?<=GPSLatitude\":).*?(?=,)') + 0.0, NULL);
```
The IGNORE statement results from the photos with malformed GPSLatitude/Longitude values taken by some devices (my old HTC phone treats it as string type instead of rational64u when the value is close to 0).

In my test bed with 13k photos, The webpage does not slow down after applying the geometric filter.

I did not implement the marker clustering feature on the map yet, since client-side clustering requires all marker information sent to the frontend and it could cause bad performance (as discussed in #121 ). I didn't find any server-side clustering library for PHP yet, so I think I should implement this feature in the next step.

BTW, My VSCode linter doesn't seem to match your coding style. Could you provide such configurations in README? Thanks a lot.


lib/Controller/LocationsController.php
```diff
@@ -0,0 +1,299 @@
...
+class LocationsController extends ApiBase
+{
+    use FoldersTrait;
+
+    /**
+     * @NoAdminRequired
+     *
+     * @PublicPage
+     */
+    public function daysWithBounds(string $minLat, string $maxLat, string $minLng, string $maxLng): JSONResponse
```
pulsejet, line 39: This code is duplicating the Days code. It should programmed as a transform instead.

lib/Db/TimelineQueryDays.php
```diff
@@ -124,6 +124,63 @@ public function getDays(
...
+     * @param bool         $archive         Whether to get the days only from the archive folder
+     * @param array        $queryTransforms An array of query transforms to apply to the query
+     * @param string       $minLat          The minimum latitude
+     * @param string       $maxLat          The maximum latitude
+     * @param string       $minLng          The minimum longitude
+     * @param string       $maxLng          The maximum longitude
+     *
+     * @return array The days response
+     */
+    public function getDaysWithBounds(
```
pulsejet, line 141: Again, this needs to be a transform, not duplicate.

pulsejet (changes requested):
Hi @RaymondHuang210129, sorry for the delay in reviewing this.

> I added another page called 'Locations' which shows the map on the top, and the timeline shows the photos taken within the visible range in the map.

Overall this seems like a good start for implementing the maps feature. I'd like to note here that this will have to be hidden behind a "experimental" flag until the rest of the features are completed: otherwise it'll be confusing UX.

> I guess the only way to change schema automatically is to have the app updated and run the migration script, so I manually added two columns (latitude, longitude) to the table `oc_memories` and runs the below query to fill each row:
> 
> ```sql
> ALTER TABLE oc_memories ADD latitude double DEFAULT NULL;
> ALTER TABLE oc_memories ADD longitude double DEFAULT NULL;
> UPDATE IGNORE oc_memories SET longitude = IF(exif REGEXP '(?<=GPSLatitude\":).*?(?=,)' AND exif REGEXP '(?<=GPSLongitude\":).*?(?=,)', REGEXP_SUBSTR(exif, '(?<=GPSLongitude\":).*?(?=,)') + 0.0, NULL);
> UPDATE IGNORE oc_memories SET latitude = IF(exif REGEXP '(?<=GPSLatitude\":).*?(?=,)' AND exif REGEXP '(?<=GPSLongitude\":).*?(?=,)', REGEXP_SUBSTR(exif, '(?<=GPSLatitude\":).*?(?=,)') + 0.0, NULL);
> ```
> 
> The IGNORE statement results from the photos with malformed GPSLatitude/Longitude values taken by some devices (my old HTC phone treats it as string type instead of rational64u when the value is close to 0).

I would not worry about this. Asking users to re-run indexing after a major upgrade is fine.

> In my test bed with 13k photos, The webpage does not slow down after applying the geometric filter.

I don't expect it to. One thing to check would be the query response times (compared to the plain timeline queries). If everything is okay they should be very similar.

> I did not implement the marker clustering feature on the map yet, since client-side clustering requires all marker information sent to the frontend and it could cause bad performance (as discussed in #121 ). I didn't find any server-side clustering library for PHP yet, so I think I should implement this feature in the next step.

Sounds good.

> BTW, My VSCode linter doesn't seem to match your coding style. Could you provide such configurations in README? Thanks a lot.

Just use Prettier for Vue and TS. For PHP, you need to run `make php-lint`

lib/Controller/LocationsController.php
RaymondHuang210129: Could you explain more about programmed as a transform? Do you mean polymorphism, or is this a language feature specific to PHP?

RaymondHuang210129:
Refactored the code as a transform filter.

pulsejet:
@RaymondHuang210129 I should've dropped a note earlier: I'm working on something similar right now for map clustering. Thankfully I think there's not a lot of conflict yet so I'm going to try merging in your code locally and continue from there. But I'll suggest putting this on hold for the time being to avoid duplicate efforts.

RaymondHuang210129:
@pulsejet Got it, thanks! I will wait for your merging process. Sorry for knowing this message late.

Issue #377, opened 2023-01-26
[Feature Request] Sort albums by name


IndrekHaav:
Looks like right now albums are sorted by creation date, most recent to oldest.

I think it would make more sense to sort them by name. This would make finding a specific album easier once there's a lot of them. Also, with some attention to the naming scheme, one could have related albums together, chronologically (start album name with year), thematically (start album name with a keyword like "Trip", "Birthday", "Wedding", etc), or something else. Current sort order affords no such control.

Hopefully this is something that can be changed and isn't fixed by the Photos app.

pulsejet:
![image](https://user-images.githubusercontent.com/10709794/223543857-079f5e66-1366-489c-b92b-b864da1d97f9.png)

PR #386, opened 2023-01-31
Sort albums by name rather than creation time

IndrekHaav:
Small change to load albums by name. As reasoned in #377, that seems like a more useful sort order than creation time.


pulsejet:
This needs a UI or settings option.

IndrekHaav:
Fair enough.

Issue #180, opened 2022-11-10
Feature Request - Mobile Device App/Wrapper


Cebrain:
Hey,

i only found memories recently and it is really great. I was exactly looking for such solution.
The only feature i would really like to see, would be an mobile app to use memories. As i was searching, i couldnt find one?

I think a wrapper for the web page would be already enough.
The problem with the browser access on mobile is that not such tech savy people will experience problems.
Example, the back button closes sometimes the complete browser instead of going back. (Think this happens how nextcloud web is build) or picking a picture with a long press, opens the browser menu infront of the drop down menu.

karpiq24:
Web app can be installed as an app, it works pretty well but I also have issues with back button closing app instead of going back.

pulsejet:
> Web app can be installed as an app, it works pretty well but I also have issues with back button closing app instead of going back.

This should not be happening. When does this happen? Also I'm assuming you're on the latest version.

karpiq24:
I just tested this and the back buttony works correctly every time. I guess it was happening on older version. Only one thing still mildly annoys me. I would love to be able to close navigation menu by clicking on the right side or by clicking back button.

Cebrain:
I am also not 100% sure where it was i had the problem with the back button, but i have the same feeling/problem as @karpiq24 that i would like to see that the back button closes the navigation menu.

pulsejet:
I agree the nav behavior is super annoying. In the next version, it'll close automatically on clicking something in it. Hopefully that's better.

I've thought about a mobile app, but it's more work than it'd seem at the surface. Basically to not be confusing, we also need to show the user's photos on the device that aren't uploaded yet in the same view. I've not written native mobile code for a long time, so not sure how easy/hard this is to achieve with modern APIs.

pulsejet:
> picking a picture with a long press, opens the browser menu infront of the drop down menu.

I just noticed this comment. This should _definitely_ not be happening. Long press is supposed to select a photo. What browser / device is this?

pulsejet:
Sounds like a bug with Firefox. I'll triage this later.

dpeley:
I'd like to chime in just to +1 this idea, I'm looking every where for a Google photos replacement on android and have yet to find anything.

IMHO, if someone is able to create a decent looking android app that can link to your server app, it would bring a LOT more attention to your brilliant work for increased usability across the board.

natrius:
[...]

PR #387, opened 2023-01-31
make endpoint accessible for external yet authenticated clients

thrillfall:
I always envied iphone users for the "Memories" feature of the Photos app . Then i found your nextcloud app. So thanks for that!
Then i started an android app that (so far) only replicates the show-"on this day n years ago"-photos: https://github.com/thrillfall/android-nextcloud-memories
It is just a proof-of-concept (code is still some big spaghetti-mud-ball) and is simply displays all thumbnails in a plain 2-column view. 

Once this has a nice user interface and experience i'd like to publish this on F-Droid. Would it be ok if this android app was named "Nextcloud Memories"? Thus being explicitly coupled to your nextcloud app?

My android/kotlin is still in newbi stage and i'd appreciate any help and additional functionality contributions.
What are your thought on this?

If you want to give it a try you'd have to apply the changes from this PR to your nextcloud instance and could use this build: https://github.com/thrillfall/android-nextcloud-memories/blob/653e2e16563576de5c95e3d7ab9272feb2592ae4/nextcloud_memories_debug.apk

### DIFF
--- a/lib/Controller/DaysController.php
+++ b/lib/Controller/DaysController.php
@@ -159,8 +159,7 @@ class DaysController extends ApiBase
 
     /**
      * @NoAdminRequired
-     *
-     * @PublicPage
+     * @NoCSRFRequired
      */
     public function dayPost(): JSONResponse
     {


pulsejet:
> I always envied iphone users for the "Memories" feature of the Photos app . Then i found your nextcloud app. So thanks for that! Then i started an android app that (so far) only replicates the show-"on this day n years ago"-photos: https://github.com/thrillfall/android-nextcloud-memories It is just a proof-of-concept (code is still some big spaghetti-mud-ball) and is simply displays all thumbnails in a plain 2-column view.

There's some reasoning in #180 of why there isn't an app yet.

Basically, the timeline and scroller code is very complex, even when implemented with a lot of help from Vue (the fact that no other FOSS photo app has this is testament to this). As a result, I've tried to refrain from rewriting everything to native code to prevent maintaining three different code bases that do the same thing. It's just a _lot_ of maintenance overhead.

So the current idea was to wait until memories is stable enough, then explore wrapping it in a native app (with some added functionality like showing the user's local photos). Of course, I'd be happy if you're willing to build / maintain a native app that integrates with memories; just that I don't have enough time to work on another project currently (there are already too many outstanding issues / missing features in the web version).

> Once this has a nice user interface and experience i'd like to publish this on F-Droid. Would it be ok if this android app was named "Nextcloud Memories"? Thus being explicitly coupled to your nextcloud app?

I'd worry that this would make error reporting even more confusing than it already is (people reporting issues with Photos here for example) ... maybe just need some indication for the separation.

> My android/kotlin is still in newbi stage and i'd appreciate any help and additional functionality contributions. What are your thought on this?

As I mentioned, I'm just constrained by time and wary of future maintenance.

> If you want to give it a try you'd have to apply the changes from this PR to your nextcloud instance and could use this build: https://github.com/thrillfall/android-nextcloud-memories/blob/653e2e16563576de5c95e3d7ab9272feb2592ae4/nextcloud_memories_debug.apk

Will give it a spin tonight :)

lib/Controller/DaysController.php
```diff
@@ -159,8 +159,7 @@ public function day(string $id): JSONResponse
 
     /**
      * @NoAdminRequired
-     *
-     * @PublicPage
+     * @NoCSRFRequired
```
pulsejet, line 162: This endpoint doesn't change state, so this wouldn't make much difference. But this definitely isn't the right way to do this: the app should be sending the CSRF token instead. This might work as a quick fix for now but there'll be trouble with other endpoints which do change state (can't remove CSRF protection on those)

thrillfall: OK, then. I'll find out how to fetch a Csrf token and sent it along

thrillfall:
I'll come up with another name then :-)

I mostly wanted to share my idea with you to check if this is going into the right direction.

I'll let you know when I reach a state I feel confident to release this to fdroid.

Thanks for your response and the idea with accessing local photos first!

thrillfall:
Closed as the android app should be working without modifying access restrictions.

PR #395, opened 2023-02-06
Support for place list with reverse geocoding

pulsejet:
This will likely conflict/merge with #376

image


RaymondHuang210129:
I believe this approach is better than #376 . Thanks for having this feature available.

pulsejet:
> I believe this approach is better than #376 . Thanks for having this feature available.

They're somewhat orthogonal (at least the implementation). We still need the map regardless.

RaymondHuang210129:
The `places-setup` command seems to break the PHP memory limit during the execution. 
My testing environment has PHP version 8.1.14 and a memory limit of 512MB.
```
Inserted 25000 places, 320.5/s, ETA: 1904s, Last: Osborne
Inserted 25500 places, 322.8/s, ETA: 1889s, Last: Stanford
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/lib/private/DB/BacktraceDebugStack.php on line 31
```

pulsejet:
Try 2GB
```bash
PHP_MEMORY_LIMIT=2G php occ memories:places-setup
```

thrillfall:
i am just running the index and can already see my holiday locations in the tiles and i just wanted to say: THANK YOU!!! This is so awesome!! :heart_decoration:

PR #397, opened 2023-02-09
Fix timezone stripping when parsing exif dates

AndreAle94:
Hello, I encountered some issues while importing videos where creation dates are stored with the following format:

immagine

As you can see, in some cases the timezone is stored using a different pattern.
This pull request is enhancing the current logic that strip the timezone to also take in consideration this case.

### DIFF
--- a/lib/Exif.php
+++ b/lib/Exif.php
@@ -180,26 +180,16 @@ class Exif
      */
     public static function getDateTaken(File &$file, array &$exif)
     {
+        // Try to parse the date from exif metadata
         $dt = $exif['DateTimeOriginal'] ?? null;
-        if (!isset($dt) || empty($dt)) {
-            $dt = $exif['CreateDate'] ?? null;
-        }
-
-        // Check if found something
         try {
             return self::parseExifDate($dt);
         } catch (\Exception $ex) {
         } catch (\ValueError $ex) {
         }
 
-        // Fall back to creation time
-        $dateTaken = $file->getCreationTime();
-
         // Fall back to modification time
-        if (0 === $dateTaken) {
-            $dateTaken = $file->getMtime();
-        }
-
+        $dateTaken = $file->getMtime();
         return self::forgetTimezone($dateTaken);
     }
 


pulsejet:
Thanks. Actually, reading CreateDate is pointless to begin with, so we can just get rid of this line:
https://github.com/pulsejet/memories/blob/0656368975b990eea208b16fcfb6ffb87d410c3a/lib/Exif.php#L185

AndreAle94:
Honestly I would leave this line as a fallback since not all media files (mainly videos like in this case) are guaranteed to have a 'DateTimeOriginal' tag

pulsejet:
There's already a fallback later which doesn't need parsing
https://github.com/pulsejet/memories/blob/0656368975b990eea208b16fcfb6ffb87d410c3a/lib/Exif.php#L200

AndreAle94:
Not sure it will return the same result: it will try to extract the creation time of the file and then the modification time as a last fallback. In my case, the file have the following timestamps in inode info:

immagine

In this case, I'm pretty sure it will pickup the creation time (2023/01/21) which is wrong

pulsejet:
`File::getCreationTime` returns the upload time to Nextcloud rather than the create timestamp (which is filesystem dependent, and might not even exist).

In hindsight, this might be more trouble, so probably best to remove that fallback too, and just rely on modified date if `DateTimeOriginal` doesn't exist (exiftool returns this anyway, AFAIK).

https://github.com/pulsejet/memories/blob/0656368975b990eea208b16fcfb6ffb87d410c3a/lib/Exif.php#L196

AndreAle94:
Yes, makes sense, thanks for the explanation! I'll update the PR

pulsejet:
Nice and simple 💯 
I'll fix the lint. Make sure to run `make php-lint` henceforth ;)

pulsejet:
Merged with https://github.com/pulsejet/memories/commit/0e345b1592b758950999c6f695a885ee27c39b79

pulsejet:
So it looks like we do need to use the `CreateDate` for certain Apple videos (the `QuickTimeUTC` option decodes the `CreateDate` from elsewhere, I'm guessing). Regardless we still don't care about the upload time.

So I'm going to revert this change and apply your original patch while removing the fallback to creation time.

PR #579, opened 2023-04-17
Escaping issues with single quotes

roliverio:
The string has a escaping issue as it breaks on the first single quotes, i've opened this PR to propose the solution enclosed, or, alternatively, use double quotes which wouldn't require escaping. Which of course causes rejection of this PR, in any case, this is to ensure proper translation string gets thru in the next nextcloud-bot run.

Thanks for your consideration.

### DIFF
--- a/src/Admin.vue
+++ b/src/Admin.vue
@@ -244,7 +244,7 @@
       {{
         t(
           "memories",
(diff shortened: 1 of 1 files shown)


pulsejet:
The code formatter automatically removes the escape, so this will be trouble

PR #602, opened 2023-04-23
Add x-msvideo mime type

dpjl:
Videos from my (old) Powershot A80 (avi files) are not processed by memories because their mime type is 'video/x-msvideo'.
Would it be possible to add this mime type by default ?
Thank you very much for your work.

David

Edit: Just a few more information: 

I tested it after adding the type in Application.php and it works perfectly.
Here is an example of these video files:
[MVI_4254.zip](https://github.com/pulsejet/memories/files/11302932/MVI_4254.zip)


pulsejet:
Awesome. 2x thanks for providing a sample!

PR #606, opened 2023-04-23
Avoid showing "Memories: Indexing process stopped before completion. Will continue on next run" as a warning

### LINKED ISSUE #605: Avoid showing "Memories: Indexing process stopped before completion. Will continue on next run" as a warning
I don't think this is exactly a bug but neither is a feature request, and I think it is closer to a bug since it is an undesirable behavior.

**Describe the bug**
I get a lot of log messages with level Warning like this:

```
Memories: Indexing process stopped before completion. Will continue on next run
```

But the cron job not finishing indexing is just normal operation, it shouldn't be a warning as there is nothing I can do to fix or improve anything.

**To Reproduce**
1. Install memories
2. Enable option to index as a cronjob
3. Add many images, more than what memories can index in one cron run

**Screenshots**
Not applicable

**Platform (please complete the following information):**

- OS: Not relevant
- Browser: Not relevant
- Version: 5.0.5

**Additional context**
Nothing relevant

**Important: Are all of the following are true?**

- I have followed the [configuration steps](https://github.com/pulsejet/memories/wiki/Configuration).
- I am NOT using `OCCWeb` to run `occ` commands
- The problem persists after clearing browser cache

**Yes**

llucax:
An indexing cron job not finishing is just normal operation, so we shouldn't emit a warning for it. This commit emits a info instead when this happens.

While we are at it, we also emit a info message if the indexing was successful.

Fixes #605.

### DIFF
--- a/lib/Cron/IndexJob.php
+++ b/lib/Cron/IndexJob.php
@@ -63,7 +63,7 @@ class IndexJob extends TimedJob
             $this->indexAllUsers();
             $this->log('Indexing completed successfully', 'success');
         } catch (Service\ProcessClosedException $e) {
-            $this->log('Indexing process stopped before completion. Will continue on next run', 'warning');
+            $this->log('Indexing process stopped before completion. Will continue on next run', 'info');
         } finally {
             \OCA\Memories\Exif::closeStaticExiftoolProc();
         }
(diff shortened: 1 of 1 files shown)


pulsejet:
Fair enough.

Issue #662, opened 2023-05-25
Always Load Full Size Image does not work in public link


rhyst:
**Describe the bug**

I have the "Always load full size image" setting on. In my folders and albums this correctly loads the full image as soon as I click on them. If I create a public share link to an album or gallery and try the same thing then the full image does not load. 

**To Reproduce**
- Ensure you have toggled the "Always load full size image" setting on
- Create an folder with pictures in memories
- Create a public share link to that folder
- In another browser or private browsing visit the public share link
- Click on an image and note that it does not load the full size image

**Screenshots**

**Platform:**

- OS: Windows
- Browser: Firefox
- Memories Version: 5.1.0
- Nextcloud Version: 26.0.1
- PHP Version: 8.1.19

**Additional context**
The "Load full image on zoom" appears to work in public shares.

- Any errors in the JS console? 

```
$ is deprecated: The global jQuery is deprecated. It will be removed in a later versions without another warning. Please ship your own. [globals.js:59:15](webpack:///nextcloud/core/src/globals.js)
InvalidStateError: A mutation operation was attempted on a database that did not allow muta
```

- Any errors in the Nextcloud server logs? No

pulsejet:
This is expected, since the "always load full image" is a user-specific setting (and public links by definition have no user). It could be possible to add a global default for this, but I'd be very careful here: this could cause huge CPU load e.g. if someone is using HEIF. The better solution here is to just increase the max preview size.

rhyst:
My use case is a single user instance where I occasionally share links with a friend to share photos from a trip or something so for me the global setting would be useful. 

If its not something you want to add, would it be possible for me to patch my own instance to do this? I am comfortable with php and js so would be happy to do this if you could point me at the relevant files.

I'm curious as to what about the setting causes CPU load? I would expect it to replace the preview link with a link directly to the original file which doesn't seem like something that should cause the server issues. But I am probably missing a lot.

rhyst:
Also is this something you would accept a PR for?

pulsejet:
> If its not something you want to add, would it be possible for me to patch my own instance to do this? I am comfortable with php and js so would be happy to do this if you could point me at the relevant files.

`src/viewer/PsImage.ts`. Or just change `highSrcCond` in `Viewer.vue`.
 
> I'm curious as to what about the setting causes CPU load? I would expect it to replace the preview link with a link directly to the original file which doesn't seem like something that should cause the server issues. But I am probably missing a lot.

[...]

PR #672, opened 2023-05-31
Add configuration for full res on zoom and full res always for public shares

rhyst:
Add seperate full res configuration for public shares to the admin settings. 

Fixes: https://github.com/pulsejet/memories/issues/662

### DIFF
--- a/src/components/Settings.vue
+++ b/src/components/Settings.vue
@@ -44,11 +44,22 @@
         >
           {{ t('memories', 'Show past photos on top of timeline') }}
         </NcCheckboxRadioSwitch>
+      </NcAppSettingsSection>
+
+      <NcAppSettingsSection id="preview-settings" :title="t('memories', 'Previews')">
+        <NcCheckboxRadioSwitch
+          :checked.sync="config.override_global_full_res"
+          @update:checked="updateOverrideGlobalFullRes"
+          type="switch"
+        >
+          {{ t('memories', 'Override the global settings for loading full size images') }}
+        </NcCheckboxRadioSwitch>
 
         <NcCheckboxRadioSwitch
           :checked.sync="config.full_res_on_zoom"
           @update:checked="updateFullResOnZoom"
           type="switch"
+          :disabled="!config.override_global_full_res"
         >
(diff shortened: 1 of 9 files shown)


rhyst:
I did start implementing it that way but then thought that a separate setting for public links was nice. i.e. I can have always on loading but public links that may be handed out to many viewers can be turned off. Or something like that.

Completely happy to change it if you would prefer.

pulsejet:
> I did start implementing it that way but then thought that a separate setting for public links was nice. i.e. I can have always on loading but public links that may be handed out to many viewers can be turned off. Or something like that.

That's the default behavior. The case I'm thinking about is the opposite: if public links have full res turned on by default, then logged in users probably want this on by default anyway (with the option of turning it off).

rhyst:
In latest commit I have made the admin settings global and added an explicit "Override" setting for the user settings. Let me know if you can think of a more elegant way.

src/components/Settings.vue
```diff
@@ -44,18 +44,30 @@
...
         </NcCheckboxRadioSwitch>
+      </NcAppSettingsSection>
+
+      <NcAppSettingsSection id="preview-settings" :title="t('memories', 'Previews')">
+        <NcCheckboxRadioSwitch
+          :checked.sync="config.override_global_full_res"
+          @update:checked="updateOverrideGlobalFullRes"
+          type="switch"
+        >
+          {{ t('memories', 'Override the global settings for loading full size images') }}
```
pulsejet, line 55: Don't think we need this; seems too complex for the average user. Just a default provided by the admin is good enough.

Anyway, this looks good overall, will make some adjustments before merging.

rhyst: I added this because I wasn't sure how otherwise a user could turn off the default, but I assume you are thinking that if an admin has turned it on then there's not a good reason that an individual user would need to turn it off?

Also would you like me to remove this or will you be doing it?

rhyst:
@pulsejet Do you need me to do anything for this to be merged?

pulsejet:
Merged, thanks! 🎉

PR #680, opened 2023-06-05
Added retry function if folder is locked and single API call 

theronakpatel:
Solving issue : https://github.com/pulsejet/memories/issues/681


theronakpatel:
Solving issue : https://github.com/pulsejet/memories/issues/681

PR #743, opened 2023-07-16
docs: Clarify that users should modify rather than replace the enabledPreviewProviders array

universalhandle:


docs/file-types.md
```diff
@@ -15,40 +17,43 @@ GIF (image/gif)
...
-## HEIC and TIFF
+... or...
+
+```php
+  'OC\Preview\PNG',
+  'OC\Preview\JPEG',
+  'OC\Preview\GIF',
+  'OC\Preview\BMP',
+  'OC\Preview\XBitmap',
+```
```
universalhandle, line 34: I added this because my config actually lists the file types individually. Not sure if that's a byproduct of being on a newer/older installation than you, but it seemed worth including.

docs/file-types.md
```diff
@@ -4,7 +4,9 @@ description: Steps to configure support for different image and video formats
 
 # File type support
 
-Memories supports the file types supported by the Nextcloud previews app. If you add support for any one of the file types below, you must run `occ memories:index` to index these files.
+Memories supports the file types supported by the Nextcloud previews app. File type support is determined in part by the values listed in the `enabledPreviewProviders` configuration parameter in your configuration file. If your `config.php` does not contain an `enabledPreviewProviders` array, this means you are using Nextcloud's defaults. Copy the array over from `config.sample.php` before adding any of the values below, or else you will effectively disable all of the defaults.
```
universalhandle, line 7: This might be obvious to many, but my `config.php` didn't include this parameter, and I didn't know there were defaults working behind the scenes. When I copy/pasted the config to add video previews I disabled everything else, and it took more than a minute to realize what I had done.

docs/file-types.md
pulsejet: Reverted this one since it's [pointless](https://github.com/nextcloud/server/blob/master/lib/private/Preview/JPEG.php) and (academically) would hurt performance. Thanks for the PR!

PR #752, opened 2023-07-28
UX improvements: batch adding to albums

### LINKED ISSUE #738: UX improvements for adding to albums process

I'm in the process of moving all of my pictures to nextcloud, but it's hard to sort them by albums.
I'm always frustrated when I:
* Try to add a picture to several albums
* Try to add a picture to album while it's opened (see fig. 1, fig. 2)
* Try to see which albums contain this particular picture

It would be nice if:
* There is a possibility to choose several albums where to add picture (or pictures) to
* There's an 'Add to album' button in sidebar
* There's a list of albums that contain this picture with possibility to remove it from album and revert this action

**Additional context**

fig. 1
image

fig. 2
image

johnSamilin:
fixes #738 
This PR enables you to select several albums in "Add to album" modal. It also brings possibility to remove a photo from albums in this modal.
image1

There's additional button in actions menu for convenience.
image2

Finally, there's new sidebar section with info about all the albums that currently contain selected photo.
image3


pulsejet:
This looks great, thanks @johnSamilin!

I've been a very busy lately so it may take a bit of time before I can take a look in detail. Before that, a couple of quick comments:
1. While I like the idea of adding and removing from multiple albums at the same time, this needs to be an option (I'm assuming it isn't right now; just ignore otherwise). An intuitive UX, for example, might involve adding to one album if you click on the album, but you can select multiple albums using the checkmark _if you like_. Somewhat similar to the photo selection in the main timeline: the photo opens if you click it, but a checkmark shows if you hover on it and you can then select multiple photos.
2. The albums list _needs_ to be in the metadata tab rather than having its own tab in the Nextcloud sidebar. This is better UX, but more importantly the sidebar may not always be accessible. For instance, the user can't open the sidebar for a shared album, where only the metadata tab is shown (there are also a few other cases).

Feel free to leave the PR as-is, I can refactor the above when I get to it. Especially for point 2: I've some thoughts about combining this change to also show the people from recognize in the metadata tab; this will likely need changes to the `ClustersBackend` interface.

johnSamilin:
Hey @pulsejet,
thanks for the feedback!
1. For me this would be anti-intuitive UX, because I think of this component as of checkbox with the label rather than of a button.
Screenshot_20230729-222122-01
Imagine if forms are submitted when you click on the label. But if you insist, I'll change that. There might be some kind of a setting in admin panel that would toggle between two approaches (if that's possible. I've just started using nextcloud recently)
2. Got it. I'll move it!

johnSamilin:
Btw, I'm not sure how do you manage translations. Is it some kind of service? I didn't add translations to new phrases like "and remove from x albums"

pulsejet:
Heads up: I'm gonna try merging this today, so will suggest holding off further changes

pulsejet:
image

pulsejet:
Thanks again @johnSamilin!

johnSamilin:
I'm glad that I could help! :)

ChildLearningClub:
@johnSamilin Just wanted to say thank you as well. 

I had almost 1000 images that had to be added to an album and was not looking forward to selecting each individually. Updated to 5.3.0-aplha.1 with your changes and with the select shift select another grabbed about half, looks like there is a ~450 selection limit per shift select , grabbed the rest and added them to the album. :)

Also really like the Albums icons under the info tab. and clicking it takes you to the album!  Really great work!!

PR #767, opened 2023-08-08
Memories metadata: Add support to Face Recognition

matiasdelellis:
I find the new approach interesting, and I want to use it.

Also I am considering hiding my People sidebar inside memories.

To minimize changes, still prioritize Recognize (At least until we're on equal terms), except when we're on the Face Recognition filter.

### DIFF
--- a/src/components/Metadata.vue
+++ b/src/components/Metadata.vue
@@ -352,7 +352,10 @@ export default defineComponent({
     },
 
     people(): IFace[] {
-      return this.baseInfo?.clusters?.recognize ?? [];
+      if (this.routeIsFaceRecognition)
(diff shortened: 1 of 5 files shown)


src/components/Metadata.vue
```diff
@@ -352,7 +352,10 @@ export default defineComponent({
     },
 
     people(): IFace[] {
-      return this.baseInfo?.clusters?.recognize ?? [];
+      if (this.routeIsFaceRecognition)
```
pulsejet, line 355: There's a missing case here: if recognize is disabled then we want to use facerecognition by default

pulsejet (approved):
Looks great, thank you!

I'll fix the minor comment before merge.

Issue #773, opened 2023-08-15
Recognize: Unassigned faces have duplicates of each image with the same square drawn


prurigro:
**Describe the bug**

In the unassigned faces section of the "People" category, each image is present multiple times for each person. Images with a bunch of people will have 30+ copies, with each of the people highlighted a bunch of times. You can't select the same image twice when assigning a particular person, so to clear them you need to select one of the photos, choose the person, select the next photo, choose the same person, etc.

I'd initially created an [issue in the recognize repo](https://github.com/nextcloud/recognize/issues/923), and one of the contributors there confirmed that the issue was present for them, but that it only happens in memories and not the same (currently unreleased) unassigned faces feature in the photos app.

**To Reproduce**

I'm not exactly sure what factors are needed to reproduce, but a number of people in the issue I linked were observing the same issue (including one of the recognize devs).

**Screenshots**

![image](https://github.com/pulsejet/memories/assets/1149238/e99041d9-6dd1-4bb1-ac35-e7496573d3fa)

**Platform:**

- OS: Linux, Windows, MacOS
- Browser: Firefox
- Memories Version: 5.2.1
- Nextcloud Version: 27.0.2
- PHP Version: 8.1.21

pulsejet:
To anyone experiencing this issue, you can apply this patch until v5.3 is out: https://github.com/pulsejet/memories/commit/b513650892b241ac4ee6b3960a90a1b26ef515a5.patch

prurigro:
Amazing- that did the trick! Thanks so much!

PR #774, opened 2023-08-16
fix: Fix a bug that caused some unassigned face photos to be unselectable

### LINKED ISSUE #721: unassigned face > selection problem
**Describe the bug**
I have a selection problem when I start to have selections, some photos are no longer selectable. 

Also, when I assign a group of photos (e.g. 10), sometimes some photos are not assigned.)

**To Reproduce**

**Screenshots**

https://github.com/pulsejet/memories/assets/45015615/8c572d86-bda4-4e6b-94a1-d0442a0ba3ce

https://discord.com/channels/1100157156742926406/1100158171621904567/1125686985063944193

**Platform:**

- OS: OMV 6
- Browser: Firefox
- Memories Version: 5.2.1
- Nextcloud Version: 27.0.0
- PHP Version: 8.2.7

**Additional context**

- Any errors in the JS console?  No
- Any errors in the Nextcloud server logs? No

g3n35i5:
This resolves #721

### DIFF
--- a/src/components/SelectionManager.vue
+++ b/src/components/SelectionManager.vue
@@ -71,9 +71,7 @@ import AlbumsIcon from 'vue-material-design-icons/ImageAlbum.vue';
 import AlbumRemoveIcon from 'vue-material-design-icons/BookRemove.vue';
 import FolderMoveIcon from 'vue-material-design-icons/FolderMove.vue';
 
-import { IDay, IHeadRow, IPhoto, IRow, IRowType, ISelectionAction } from '../types';
-
-type Selection = Map<number, IPhoto>;
+import { IDay, IHeadRow, IPhoto, IRow, IRowType, ISelectionAction, SelectionProxy } from '../types';
 
 /**
  * The distance for which the touch selection is clamped.
@@ -124,14 +122,14 @@ export default defineComponent({
   data: () => ({
     show: false,
     size: 0,
-    selection: new Map<number, IPhoto>(),
+    selection: new SelectionProxy(),
     defaultActions: null! as ISelectionAction[],
 
     touchAnchor: null as IPhoto | null,
     prevTouch: null as Touch | null,
     touchTimer: 0,
     touchMoved: false,
-    touchPrevSel: null as Selection | null,
+    touchPrevSel: null as SelectionProxy | null,
     prevOver: null as IPhoto | null,
     touchScrollInterval: 0,
     touchScrollDelta: 0,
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -228,7 +283,7 @@ export type ISelectionAction = {
   /** Icon component */
   icon: any;
   /** Action to perform */
-  callback: (selection: Map<number, IPhoto>) => Promise<void>;
+  callback: (selection: any) => Promise<void>;
```
g3n35i5, line 286: The type should be `SelectionProxy` but this doesn't work. Can someone help out here?

src/components/SelectionManager.vue
```diff
@@ -246,8 +244,17 @@ export default defineComponent({
...
+    deleteSelectedPhotosById(delIds: number[], selection: SelectionProxy) {
+      let photosToDelete: IPhoto[] = []
+      for (const photo of selection.values()) {
+        if (photo === undefined) {
+          continue
+        }
+        if (delIds.indexOf(photo.fileid) > -1) {
+          photosToDelete.push(photo)
+        }
+      }
```
pulsejet, line 256: 1. Why not just `selection.values().filter( ... )`?
2. Can get a safer time complexity by making `delIds` a `Set` first

src/components/SelectionManager.vue
```diff
@@ -563,19 +570,19 @@ export default defineComponent({
...
         this.selectionChanged();
       } else {
         photo.flag &= ~this.c.FLAG_SELECTED;
 
         // Only do this if the photo in the selection set is this one.
         // The problem arises when there are duplicates (e.g. face rect)
         // in the list, which creates an inconsistent state if we do this.
-        if (this.selection.get(photo.fileid) === photo) {
-          this.selection.delete(photo.fileid);
+        if (this.selection.get(photo) === photo) {
```
pulsejet, line 584: This comment and check should no longer be relevant.

src/types.ts
```diff
@@ -220,6 +220,61 @@ export type ITick = {
   key?: number;
 };
 
+export class SelectionProxy {
```
pulsejet, line 223: `types.ts` is intended only for interfaces, not implementations (this way, you can `import type ...`). This class can move to another file.

src/types.ts
```diff
@@ -220,6 +220,61 @@ export type ITick = {
...
+  private map: Map<string, IPhoto> = new Map<string, IPhoto>();
+
+  add(photo: IPhoto): this {
+    this.map.set(photo.key!, photo);
+    return this;
+  }
+
+  get(photo: IPhoto): IPhoto | undefined {
+    return this.map.get(photo.key!);
```
pulsejet, line 232: This, precisely, could be the issue. `Photo.key` is nullable because, well, it can be null. The bang operator discards this information without a check. There are indeed a bunch of places where we "fake" a `IPhoto` object that doesn't have a key.

By now though, I'm hoping the selection manager doesn't have to deal with these things. So a runtime assertion here _might_ suffice, so we know if something breaks.

pulsejet (commented):
Thanks for this @g3n35i5. I'll go ahead and merge this in with a few updates today.

BTW #773 is related to aggravating this issue; that patch might help.

PR #788, opened 2023-08-23
Update memories-vs-photos.md

meichthys:
Add note about `draw` and `Shared with you` features missing in memories.


pulsejet:
Yes, good point.

PR #803, opened 2023-09-01
Add per-group and skip cleanup options

akhil1508:
- For tens of thousands of users on an instance, running the index command is a must before enabling the app
- Sometimes,  need to re-run for a particular group so added `group`  option
- Also cleanup itself takes a few minutes and when running per-user, it wastes a lot of time, so added `skip-cleanup` option that defaults to `false`

### DIFF
--- a/lib/Command/Index.php
+++ b/lib/Command/Index.php
@@ -29,7 +29,6 @@ use OCP\Files\IRootFolder;
 use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IUser;
-use OCP\IGroup;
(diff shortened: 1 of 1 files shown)


akhil1508:
@pulsejet FYI, let me know if these changes are good :)

pulsejet:
Thanks @akhil1508. BTW you can run `make php-lint` after `make dev-setup` to fix lint issues in the future.

> For tens of thousands of users on an instance, running the index command is a must before enabling the app

Are you really running this with this many users? I'm curious about performance; never tested it beyond two users myself.

Issue #765, opened 2023-08-05
Heatmap instead of Tiles in Map View


reox:
**Is your feature request related to a problem? Please describe.**

The Google Photos app on Android shows the map as a heatmap instead of tiles with the number of images in that region.
Especially if you zoom out very far, i.e., to cover whole Europe, the heatmap gives a better indication than the (relatively large) tiles.

**Describe the solution you'd like**
Also implement a heatmap for the map feature.

**Describe alternatives you've considered**
There are not many ways to show density on a map.

Another neat way might be to count the number of images per country (or district`*`, if zoomed in) and then colorize the countries by the number of pictures taken there. However, this might get funny when you took pictures off-shore...

**Additional context**

Here is an image how it could look like: https://stackoverflow.com/questions/28952112

`*` Could you not get this information from the reverse geocoding?

tacruc:
There is a leaflet heat plugin to render this: https://github.com/Leaflet/Leaflet.heat
I used it for the maps app to render many device points. It worked nicely.
Some more information can be found in the PR https://github.com/nextcloud/maps/pull/970.

PR #807, opened 2023-09-06
[Proof of principle] Photos heat map

tacruc:
Add's a heatmap layer to show the location of the image on the map. This was suggested in #765. 

Screenshot_20230906_234018-1

### DIFF
--- a/src/components/top-matter/MapSplitMatter.vue
+++ b/src/components/top-matter/MapSplitMatter.vue
@@ -35,9 +35,9 @@
         :initial-points="points"
         :options="{
           // minOpacity: null,
-          // maxZoom: null,
-          radius: 15,
-          blur: 10,
+          maxZoom: 16,
+          radius: 50,
+          blur: 30,
           gradient: { 0.4: 'blue', 0.65: 'lime', 1: 'red' },
         }" />
     </LMap>
@@ -283,7 +283,7 @@ export default defineComponent({
       // Params have changed, quit
       const res = await axios.get(url);
       this.points = res.data.map((c) => {
-        return [c.center[0], c.center[1], c.count]
+        return [c.center[0], c.center[1], 1]
       });
     },
 
--- a/src/components/top-matter/MapSplitMatter.vue
+++ b/src/components/top-matter/MapSplitMatter.vue
@@ -30,6 +30,16 @@
           </div>
         </LIcon>
       </LMarker>
+      <LHeatMap v-if="points.length >= 50"
+        ref="heatMap"
+        :initial-points="points"
+        :options="{
+          // minOpacity: null,
+          // maxZoom: null,
+          radius: 15,
+          blur: 10,
+          gradient: { 0.4: 'blue', 0.65: 'lime', 1: 'red' },
+        }" />
     </LMap>
   </div>
 </template>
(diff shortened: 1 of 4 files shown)


src/components/top-matter/MapSplitMatter.vue
```diff
@@ -245,6 +259,34 @@ export default defineComponent({
       this.animateMarkers();
     },
 
+    async fetchPoints() {
+      const zoom = "18";
+      let minLat = -90;
+      let  maxLat = 90;
+      let minLon = -180;
+      let maxLon = 180;
```
pulsejet, line 267: This will kill performance on slower servers. It might make sense to change the granularity of what is fetched when the heatmap is enabled, but can't load everything.

src/components/top-matter/MapSplitMatter.vue
```diff
@@ -245,6 +259,34 @@ export default defineComponent({
...
+      const bounds = this.boundsToStr({ minLat, maxLat, minLon, maxLon });
+
+      // Make API call
+      const url = API.Q(API.MAP_CLUSTERS(), { bounds, zoom });
+
+      // Params have changed, quit
+      const res = await axios.get(url);
+      this.points = res.data.map((c) => {
+        return [c.center[0], c.center[1], 1]
+      });
```
pulsejet, line 287: If we are loading this current screen only, this might as well be a computed prop derived from `this.clusters`

pulsejet (changes requested):
Overall this is nice, but will need some touch up.
1. Separate user-specific settings to enable / disable the heatmap and the clusters shown over it
2. Performance needs to be benchmarked; can't load everything.
3. I don't prefer the current color scheme, it seems a bit out of place

I can take care of the above but it will take a while (out travelling for the next few weeks)

tacruc:
@pulsejet thanks for your fast response and good review. I agree with your comments and enjoy your holiday.

I would like to add some performance observations. On the devices I tested with the heatmap was fasty able to handle on the order of 10k-100k points. This is significant more then the marker plugin. Therefore the traidoff on how much data to load might be different. And it should be sufficient to cover the World, if we can get the request to be fast on small servers. 

Additional getting the points in one request makes the heatmap, work smoothly on tilling and zooming, as no data has to be loaded.

Still I agree that the request should respond fast and not load millions of points. Additional I agree that we should test on a raspberry pi server and a low-end smartphone.

On the other hand on high latancy connection like bad mobile connection one request on initialisation might be preferable to reloading on panning the map.

src/components/top-matter/MapSplitMatter.vue
tacruc: As said in the long response. From my experience the heatmap is capable to handle much more Datapoints than the markers. 

If this observation holds I think it makes sense to seperate this two data sources.

src/components/top-matter/MapSplitMatter.vue
tacruc: Yes I was thinking of implementing an own endpoint, which just delivers the point array of the appropriate size to cover the World. This response might additionally be cached for some time (Or until a certain number of images was changed.)

servusdei2018 (approved):

PR #812, opened 2023-09-10
add heif, webp, avif

JanisPlayer:
I'm currently making a change to Nextcloud that allows for the creation of JPEG previews from AVIF images.
With this change, if the proposal is accepted, this app would also be up-to-date.
In addition, I noticed that .heif and .webp were missing, so I went ahead and added them.


pulsejet:
I'm gonna go ahead and add in `image/heif` and `image/webp` since this might remain blocked by `nextcloud/server` for a while.

PR #833, opened 2023-09-26
Set default timeline path option

akhil1508:
- Would be nice if it was an admin option to set a default timeline path
- This PR attempts to set the default timeline path

### DIFF
--- a/lib/Controller/OtherController.php
+++ b/lib/Controller/OtherController.php
@@ -82,6 +82,7 @@ class OtherController extends GenericApiController
             };
 
             $defaultTimelinePath = $this->config->getSystemValue('memories.default_timeline_path', 'EMPTY');
+
             return new JSONResponse([
                 // general stuff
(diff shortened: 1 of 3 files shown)


pulsejet:
Unless I'm missing something, the default is basically never used since the user chooses their timeline paths on the first setup page. What are you trying to target here?

akhil1508:
> Unless I'm missing something, the default is basically never used since the user chooses their timeline paths on the first setup page. What are you trying to target here?

@pulsejet I have a server with many users where I would like to provide opinionated default and not have a setup screen. So this would be a way for me to provide a default path of "/" for example.

Secondly, there is also this mildly annoying error about "Photos/" path not being present when the user doesn't have this directory, this shows up on the first screen. The code here  doesn't fix it but by setting the default timeline path, I can bypass this error.

pulsejet:
Changed the option to `memories.timeline.default_path`

PR #839, opened 2023-09-29
Translate tag names

ASDFGamer:
This Pull-Request contains two changes related to translation of tag names.
The first change fixes a Problem, that the title of tag detail pages like /apps/memories/tags/Frog wasn't translated even though the name of is Tag is translated on the overview page.
The second change sorts the Tags on the overview page in the correct order for the current language. Previously this was sorted by english names and the translated names were shown.

### DIFF
--- a/src/components/ClusterView.vue
+++ b/src/components/ClusterView.vue
@@ -82,6 +82,18 @@ export default defineComponent({
   },
 
   methods: {
+    translated_tag_name(first_item: ICluster,second_item: ICluster) : number  {
+      const first_item_translated = this.t("recognize", first_item.name) 
+      const second_item_translated = this.t("recognize", second_item.name) 
+      if (first_item_translated < second_item_translated) {
+        return -1;
+      }
+      if (first_item_translated > second_item_translated) {
+        return 1;
+      }  
+      return 0;
+    },
+
     async routeChange() {
       try {
         this.items = [];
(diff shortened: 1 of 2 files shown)


src/components/top-matter/ClusterTopMatter.vue
```diff
@@ -35,7 +36,7 @@ export default defineComponent({
     name(): string | null {
       switch (this.$route.name) {
         case 'tags':
-          return this.$route.params.name;
+          return t('recognize',this.$route.params.name);
```
pulsejet, line 39: Pls use [prettier](https://prettier.io/)

src/components/top-matter/ClusterTopMatter.vue
```diff
@@ -13,6 +13,7 @@
 <script lang="ts">
 import { defineComponent } from 'vue';
 
+import { translate as t } from '@nextcloud/l10n';
```
pulsejet, line 16: No need to import this. Just `this.t(...)` works.

src/components/ClusterView.vue
```diff
@@ -82,6 +82,18 @@ export default defineComponent({
...
+      const second_item_translated = this.t("recognize", second_item.name) 
+      if (first_item_translated < second_item_translated) {
+        return -1;
+      }
+      if (first_item_translated > second_item_translated) {
+        return 1;
+      }  
+      return 0;
+    },
+
```
pulsejet, line 96: This doesn't play nice with accents, e.g. `Électronique` is after `test`

ASDFGamer:
Thanks to the fix for the Problems that you mentioned with this pull request. Just noticed that you fixed that as i was trying to fix the problems. The use of display_name in ICluster is also probably very helpfull to save resources

PR #840, opened 2023-09-29
Show number of pictures per place

ASDFGamer:
This shows the number of pictures per place.
This function was previously disabled, if the parameter 'mark'
wasn't set to 0. I found no reason why this was the case.

### DIFF
--- a/lib/ClustersBackend/PlacesBackend.php
+++ b/lib/ClustersBackend/PlacesBackend.php
@@ -162,7 +162,7 @@ class PlacesBackend extends Backend
         $lang = Util::getUserLang();
         foreach ($places as &$row) {
             $row['osm_id'] = (int) $row['osm_id'];
-            $row['count'] = $marked ? 0 : (int) $row['count']; // the count is incorrect
+            $row['count'] = (int) $row['sub.count'];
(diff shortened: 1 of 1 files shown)


lib/ClustersBackend/PlacesBackend.php
```diff
@@ -162,7 +162,7 @@ public function getClustersInternal(int $fileid = 0): array
         $lang = Util::getUserLang();
         foreach ($places as &$row) {
             $row['osm_id'] = (int) $row['osm_id'];
-            $row['count'] = $marked ? 0 : (int) $row['count']; // the count is incorrect
+            $row['count'] = (int) $row['sub.count'];
```
pulsejet, line 165: `$row['count']` not `sub.count`

pulsejet (changes requested):
Good catch. The count was incorrect in an earlier implementation but should be OK now.

lib/ClustersBackend/PlacesBackend.php
ASDFGamer: You'r right, I thougth i tested it witch just count and it didn't work, but it works now. I must have made a wrong test. It should be fixed now

Issue #324, opened 2023-01-02
[feature request] Keep subfolders structure when sharing with public link


aboulan:
Hello,
When sharing with public link a folder containing subfolders, it would be nice to keep the folder structure in the public view instead of having a flat timeline.

Cl00e9ment:
I had the same thought!
I'm currently working on it. A PR will arrive soon.

pulsejet:
Available with 6.0

PeggyFree:
Excuse me to ask, I switched from the 5.x to the 6.0.1 , I had a message confirming the successful upadte, but I still see my public shares flat instead of seeing folders.

Cl00e9ment:
There is a button on the top right of the page to switch between the timeline view and the folder view.

PeggyFree:
Thank you for your answer. The folder/timeline button toggle finally appeared. I was waiting for this, thank you all.

PR #880, opened 2023-10-19
share: show directory structure

Cl00e9ment:
Implemented feature request #324 

## Security considerations

API endpoint `/api/folders/sub` is now marked as `@PublicPage`, but I removed the `path` information from the returned data. This was the absolute path, revealing the name of folders outside the share. The web client now compute the path by itself, see bellow.

## Some points to check before merging

- If a folder `child` is in a folder `parent`, the web client assumes that its path is `{parent path}/{child name}`. Looking at the code that was present before, it seems that it may not be always the case? Maybe?
- Route `/s/{token}/authenticate/{redirect}` is in conflict with `/s/{token}/{path}`. It's only an issue if you refresh the page on a directory which parent is named `authenticate`, which parent is the share root. So it's highly unlikely that this will cause any issue, but it's cleaner if it's addressed. How can we re-organise the routes?
- Some HTTP requests end up with `412` because of a failed CFRS check, but I don't really understand this issue.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 function getWildcard($param)
 {
     return [
@@ -34,7 +36,6 @@ return [
         w(['name' => 'Page#tags', 'url' => '/tags/{name}', 'verb' => 'GET'], 'name'),
 
         // Public folder share
-        ['name' => 'Public#showShare', 'url' => '/s/{token}', 'verb' => 'GET'],
         [
             'name' => 'Public#showAuthenticate',
             'url' => '/s/{token}/authenticate/{redirect}',
@@ -45,6 +46,7 @@ return [
             'url' => '/s/{token}/authenticate/{redirect}',
             'verb' => 'POST',
         ],
+        w(['name' => 'Public#showShare', 'url' => '/s/{token}/{path}', 'verb' => 'GET'], 'path'),
 
         // Public album share
         ['name' => 'PublicAlbum#showShare', 'url' => '/a/{token}', 'verb' => 'GET'],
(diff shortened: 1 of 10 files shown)


pulsejet:
This looks **great** overall. I might just make a few minor tweaks before merging (e.g. variable names). I've also improved the sanitization logic on master which should help here (not a security issue though, since Nextcloud already sanitizes input everywhere)

> If a folder child is in a folder parent, the web client assumes that its path is {parent path}/{child name}. Looking at the code that was present before, it seems that it may not be always the case? Maybe?

We can deal with this if something comes up later.

> How can we re-organise the routes?

Hmm we can't; that route is given to us. Maybe it's better to use a query parameter for the path in this case? What do you think?

> Some HTTP requests end up with 412 because of a failed CFRS check

Weird. Does this happen sporadically?

pulsejet:
One important point: we need to make sure the folders the user sees in here are sharable (best to only get folders on the same internal storage). Shared folders, e.g. might not allow resharing.

pulsejet:
FYI I'm working on merging this; so we don't get a conflict

pulsejet:
> Route /s/{token}/authenticate/{redirect} is in conflict with /s/{token}/{path}. It's only an issue if you refresh the page on a directory which parent is named authenticate, which parent is the share root.

Left this as-is. No biggie.

> we need to make sure the folders the user sees in here are sharable

I confirmed Nextcloud doesn't let you folders with shares in them. This PR doesn't change any behavior here so we can deal with this later if needed.

Cl00e9ment:
Thank you for this fast revue/merge!

pulsejet:
Thanks to you! More contributions are always welcome 😄

PR #888, opened 2023-10-26
Make error about tool version mismatch clearer

SuperSandro2000:


lib/Service/BinExt.php
```diff
@@ -76,7 +76,7 @@ public static function testExiftool(): string
         $version = trim($out);
         $target = self::EXIFTOOL_VER;
         if (!version_compare($version, $target, '=')) {
-            throw new \Exception("version does not match {$version} <==> {$target}");
+            throw new \Exception("exiftool version does not match: expected {$target} but found {$version}");
```
pulsejet, line 79: the binary name is redundant, IIRC. The UI already shows it

SuperSandro2000: This message is also shown in occ which has no context around it.

PR #900, opened 2023-10-30
docs: Add nixos section to get automatic indexer running

Shawn8901:
It might be a good idea if another nixos user can cross-check if thats really fixing the issue.
For reference i did install memories impurely via the nextcloud app store and *not* via `services.nextcloud.extraApps`.


pulsejet:
Thanks!

Issue #558, opened 2023-04-10
Allow for more sorting options within Albums view


XTREEMMAK:
**Is your feature request related to a problem? Please describe.**
At present, within Albums view, you are currently limited to two sorting options; "By Date" and "By Name". The unfortunate side is that you can't specify in which direction those sorts take place aka Ascending or Descending (A -> Z or Z -> A). Also, soring by date is based on the Album's creating date, regardless of when those images were shot.

**Describe the solution you'd like**
I'd like to see more sorting options (Ascending, Descending, perhaps Highest Rated, By Number of Photos, etc), and possibly a custom key/value for albums so you can specify a custom date/time parameter that the sort options will use instead aka, letting the user specify the date specifics manually.

**Describe alternatives you've considered**
You can use the Browser's search option to find the album text that you're looking for, but this is definitely not very intuitive.

Rollbacke:
Hi team,
I'm taking this thread to also add the ability to save preferred sort order.
When we change the sort order, at the next connection, it always returns to "sort by date".

pulsejet:
@Rollbacke this is a local setting, it'll only persist in your browser.

Rollbacke:
Hi @pulsejet, precisely. What I propose is to add it to the settings so that the option persists outside the browser. :)

ldaverio:
> Hi @pulsejet, precisely. What I propose is to add it to the settings so that the option persists outside the browser. :)

I second that suggestion :) A reverse chronological order may make sense in a timeline, but not so much in an album, when you want people to explore your photos in the same order as they were taken :)

pulsejet:
> > Hi @pulsejet, precisely. What I propose is to add it to the settings so that the option persists outside the browser. :)
> 
> I second that suggestion :) A reverse chronological order may make sense in a timeline, but not so much in an album, when you want people to explore your photos in the same order as they were taken :)

@ldaverio albums are already sorted chronologically (not reverse), unless you change this in user settings. This issue is about the list of albums, not inside the album itself.

ldaverio:
Oh, sorry, apparently I didn't read the previous posts closely enough, all my apologies.

I'm freshly transitioning from the OwnCloud "Gallery" plugin, and up to now I haven't found how I can share a folder by public link, and have the photos sorted in chronological order. They seem to always come up in reverse order when I open the link as an anonymous user. Also, the "Download all" link seems to be missing, but that's secondary. What matters to me most, is that I can show my photos in the order I took them. 

[...]

PR #969, opened 2023-12-20
Album sorting fixed and changed - proposal

frederikb96:
Partly addresses #558
- New buttons for sorting albums in the album view in both directions (ascending and descending, both for date and name)
- This was difficult to adjust in mobile mode. Had to switch from album view to explore, than to settings and then toggle which sorting mode to use.
  - Old function was also not working since the config was not queried during the sorting process
  - Alternative would be to simply fix it by querying the config, if you think it is better to keep the toggle in the settings to keep cleaner UI
  - However, this would lead to the more difficult toggling of the sorting order
  - Moreover, there is currently no toggle in the settings to decide whether to sort albums from A-Z or Z-A
  
I am generally unsure, about the old setting, since it is called in the UI "Sort album oldest first" but in the code it is the config value named "sort_album_month". This config value, as said, is not used in the querying of the sorting order in album.ts. However, it actually appears in Timeline.vue in isMonthView(). I don't understand this. But maybe you know :) If it is not needed in the Timeline.vue, one could completly remove the old config value "sort_album_month". But maybe, I just did not understand the purpose and you had some other thing in mind :)

### DIFF
--- a/src/components/top-matter/AlbumTopMatter.vue
+++ b/src/components/top-matter/AlbumTopMatter.vue
@@ -15,25 +15,51 @@
           <SortIcon :size="20" />
         </template>
 
+        <!-- Sort by date descending -->
         <NcActionRadio
           name="sort"
-          :aria-label="t('memories', 'Sort by date')"
+          :aria-label="t('memories', 'Sort by date (descending)')"
           :checked="config.album_list_sort === 1"
           @change="changeSort(1)"
           close-after-click
         >
-          {{ t('memories', 'Sort by date') }}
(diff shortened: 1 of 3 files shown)


pulsejet:
`sort_album_month` has nothing to do with sorting the album list: it sorts the photos _inside_ the album differently.

From a quick look this seems fine; for a minor comment, I think it would make better UX to separate the ascending / descending to a separate dropdown. I expect there to be more sorting options in the future (e.g. sort by last updated or sort by latest photo), then those would also require a ascending / descending option.

pulsejet:
7.1 has an adapted version of this code :)

Issue #746, opened 2023-07-22
Enable search for albums based on their title


quentinus95:
**Is your feature request related to a problem? Please describe.**
There are a lot of open tickets regarding search, but a lot are asking for search related to tags or images content. One (maybe) low-hanging fruit would be to have search based on the albums title (in the albums page).

Let's consider you have dozens (if not hundreds) of albums and want to quickly find a specific picture in an album you already have in mind. It is then almost impossible to find it (especially on mobile where you cannot use the browser search feature when “installing Memories as an app”).

Note that this is an issue for grandma adoption.

**Describe the solution you'd like**
When using the search feature of Nextcloud, I would expect the albums list to be filtered (and/or the search results to show the matching albums).

**Describe alternatives you've considered**
Using browser search feature, but it is not doable on mobile.

As always, thank you for your awesome work on this project. If I can do any non-code contribution work to improve this feature request, please let me know.

meonkeys:
Seems like #674 covers this.

jsddsfoh:
I was happy to read "Basic search functionality" in changelog from v7.1.0.
But unfortunately it's "only" albumname and place based.
I would love to see a foldername based search, as i organize my photos in folders :)

PR #970, opened 2023-12-20
Temporary Album Search function added

frederikb96:
Addresses #746
- Temporary search function, until unified search through nextcloud search function is available
- Very useful for hundreds of albums
- I tried to implement it without changing a lot, so simply added a new config value which can be used similar to the sorting values to trigger the filtering for search queries
- There are probably better ways to solve this

I really needed this feature, so I added it in this simple form until a proper solution is available. But I understand, if you would rather not accept temporary solutions :)

### DIFF
--- a/src/components/top-matter/AlbumTopMatter.vue
+++ b/src/components/top-matter/AlbumTopMatter.vue
@@ -10,6 +10,16 @@
     <div class="name">{{ name }}</div>
 
     <div class="right-actions">
+      <!-- Search Bar-->
+      <div v-if="isAlbumList" class="search-bar-container">
+        <NcInputField
+          :value="searchQuery"
+          :placeholder="t('memories', 'Search')"
+          @input="searchQuery = $event.target.value"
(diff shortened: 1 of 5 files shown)


pulsejet:
Thanks for this. I'm on the road for a while, so probably can't get to reviewing / merging in detail until early next year.

General comments:

1. I'm not sure why we need a setting here. Just using fuse.js to filter should work fine (the add to album modal already has a search box that does this; probably can just copy over that code). We need a setting only if this has to be persisted for some reason (opening an album and clicking back doesn't count; that's just state preservation which is a separate issue throughout the app)

2. It might make sense to generalize and move this to the `ClusterView` instead. That should allow filtering tags / faces / places too with minimal extra effort.

frederikb96:
Thanks a lot for the fast reply 😊
Perfekt, thanks for the hints, I really appreciate it.  I am really new to Nextcloud App development and grateful for tips. I will improve the code as soon as I find time 

No stress with merging any of my request, since I can already use my changes on my server anyways  🙃👍

PR #1015, opened 2024-01-23
FEAT: Ignore files and folders with regex option. 

distante:
This PR introduces a control to ignore files and folders using a custom regex when enabled.

<hr>

I also included a [Visual Studio Code Dev Container](https://code.visualstudio.com/docs/devcontainers/containers) for people who want to contribute but do not have the environment for it (like me).

The container will install the node modules the first time it is created, and start watch mode on start.

Also, a Nextcloud server with the Memories App as a Custom App is being started with the container, the config and data folders are exposed inside `.devcontainer/config` and `.devcontainer/data` but ignored from git.

I haven't done any coding with PHP in a really long time so maybe some tools still need to be added. But I use it to do and test this PR.

image

<hr>

On the topic of tests. I have not found how the occ scripts are tested. I tried the playwright test but they do not have anything with indexing (also, it needs more configuration for the dev container, for now, the port has to be manually changed).  @pulsejet if tests are needed please point me in the correct direction. Thanks!


pulsejet:
Thanks for your contribution!

There are a couple of issues here:

1. Indexing files in hidden folders is intentional, this is required for archive functionality to work. These are actually hidden from the timeline by default (there might be a bug in hidden _files_)
2. Not indexing any file in the path that is supported causes issues currently, since the next cron job also attempts to index them. So this keeps making the cron jobs slower and slower. This will be fixed once there's an exclusion / failure list
3. Better to have a configurable regex pattern for ignore

pulsejet:
Hidden files are addressed in https://github.com/pulsejet/memories/commit/629f40d70c65ebe1dc073f28f73b62999cb54361

distante:
So I have to also add that check (now regex) to the cronjob, correct?

Can you point me to where is it?

Thank you!

pulsejet:
> So I have to also add that check (now regex) to the cronjob, correct?

Actually I forgot we combined both to the service, so that part is fine. In that case the two issues would be:

1/ You're filtering by file basename only, but folders get separately filtered below. So this will actually not ignore the `@Recycle` folder, for instance.

2/ The first point in my comment. Ignoring folders with a `.` at the start would break archive functionality. Also files starting with `.` are already hidden now with the commit referenced above.

My suggestion: use `array_filter` on `$nodes` instead of `$files` to match with the regex pattern. The pattern can/should also be a lot more explicit to what we're trying to do, e.g. default to `^@Recycle$` to avoid unintended consequences (even though it seems unlikely, someone might have photos starting with `@` for all we know.

distante:
@pulsejet I had added some updates, I have added a toggle to enable the pattern if wanted and a link for some checkers.

I have also added some logs so errors on wrong patterns are shown and logs for folders/files ignored.

I hope it is now Ok. 

Pd: I still have to configure the container to resolve all nextcloud/symphony types

.devcontainer/docker-compose.yml
```diff
@@ -1,14 +1,28 @@
...
+# nextcloud:
+#   name: nextcloud
+# apps:
+#   name: apps
+# config:
+#   name: config
+# data:
+#   name: data
+# theme:
+#   name: theme
```
pulsejet, line 28: Why all these comments?

.devcontainer/docker-compose.yml
```diff
@@ -1,14 +1,28 @@
...
+  nextcloud-server:
     build:
       context: .
       dockerfile: Dockerfile
     volumes:
+      # - nextcloud:/var/www/html
+      # - apps:/var/www/html/custom_apps
       - ./custom_apps:/var/www/html/custom_apps
-      - ../:/var/www/html/custom_apps/memories
+      - ./../:/var/www/html/custom_apps/memories
```
pulsejet, line 11: Is the `./` at the start required? Seems a bit odd to me.

.devcontainer/docker-compose.yml
```diff
@@ -1,14 +1,28 @@
 name: 'nextcloud-with-memories'
 services:
-  nextcloud:
+  nextcloud-server:
     build:
       context: .
       dockerfile: Dockerfile
     volumes:
+      # - nextcloud:/var/www/html
+      # - apps:/var/www/html/custom_apps
```
pulsejet, line 9: Needed?

pulsejet (commented):
I'm gonna cherry pick the two changes separately, probably make some minor changes to the config. We likely don't need two separate config options, just a single one.

pulsejet:
I've updated the dev container setup to start and set up nextcloud with mariadb. Codespaces indicates that everything should be working though I've not tried it locally.

.devcontainer/docker-compose.yml
distante: Mostly for other developers that are new to nextcloud (like me) to know that they can mount those folders if needed. Should I remove them?

pulsejet: Don't worry about it, I already picked this change. Turns out we need to use volumes anyway for it work right; folders don't play nice with ownership somehow (the server runs as www-data).

The dev container on master should start and set up a full nextcloud instance with memories now (thanks!)

.devcontainer/docker-compose.yml
distante: Mmm, I think vscode formatter add them. I do not remember that I add it. But I do not remember now to be honest.

.devcontainer/docker-compose.yml
distante: Will delete them.

distante:
So I should rebase, correct?

pulsejet:
Moved around some things, but essentially the same. Thank you!

image

distante:
Thank you @pulsejet! The last weeks I had almost no time between the work and children to do any changes.

PR #1031, opened 2024-02-03
Remove the outlines from the Memories icon

gvtulder:
This pull request removes the black outline from the Memories icon to make it look cleaner and more similar to the other Nextcloud icons.

The current version of the icon has a very thin black outline around the photos. Especially on high-resolution displays such as an iPad, this makes the icon look less clear than the standard Nextcloud icons, which do not have an outline.

This is an iPad screenshot of the old (left) and new (right) Memories logo:
screenshot-nextcloud-memories-icon-only
In the Nextcloud menu bar, the old logo on the left clearly looks different from the others:
screenshot-nextcloud-memories-icon

On other systems the change might be more subtle, because the outline is not as visible. Check [this live demo](https://www.vantulder.net/tmp/memories-icon-example) to compare the icons in a browser of your choice.


pulsejet:
Nice!

PR #1033, opened 2024-02-05
Update AndroidManifest.xml

Ulyzzeus:
adds the option to select app as media management app, so users don't have to give permission every time they want to delete a local file


pulsejet:
It's actually intentional. There's a confirmation box anyway for remote files; that's skipped if local files are present.

Croydon:
Given that Memories can also display local media files without any problem, I would like very much to be able to open Memories by default for viewing media. My understanding is that the change in this PR enables this 👍

Croydon:
It would still be nice, if Memories could be a photo viewer all around, and be recognized as such by the Android system.

pulsejet:
@Croydon yes, agree. Planning to look into this at some point.

Issue #1060, opened 2024-03-04
Move to Nextcloud Vue 8


pulsejet:
`[NC_VUE]`

- [ ] Change NcListItem slots
- [ ] Change NcListItem title to name
- [x] Add `js` suffix to component imports
- [ ] https://github.com/nextcloud-libraries/nextcloud-vue/pull/5342

PR #1073, opened 2024-03-12
Adds uploading button and uploading modal

### LINKED ISSUE #69: Make it possible to upload photos directly from the app
In the upcoming photo app 25, nextcloud get an option to  upload photos from the photo app.
It would be nice to have such option in memories also.

johnSamilin:
DRAFT

fixes #69 

- Upgraded @nextcloud/vue to 8 version
- Added UploadModal component
- Added button to sidebar
- Photos are being uploaded to the current folder (or `/` by default)

I'll appreciate any help with testing.


pulsejet:
Great! Can you please clean up your diff? You seem to have some issue with the prettier configuration (wrong version?) which makes it impossible to review this.

pulsejet:
> Upgraded @nextcloud/vue to 8 version

See #1060. I believe the patch is merged in the new version but this is still very high risk right now.

> Photos are being uploaded to the current folder (or / by default)

This will need to be configurable; definitely don't want them going to `/`

johnSamilin:
@pulsejet @nextcloud/vue 8 is a necessity because @nextcloud/upload requires it

pulsejet:
I see. Does it still work on 27 then? The target is to support at least two prior versions.

pulsejet:
@johnSamilin I'll likely carry this forward in a separate branch. Thanks for the contribution!

johnSamilin:
@pulsejet got it. I hope this PR was helpful. If not, sorry, I'm not very familiar with nextcloud and vue still

pulsejet:
Yup, I just built over this :)

ChildLearningClub:
The upload button appears when sharing the album as public link, but nowhere within the Memories settings or when creating the public link can I see the options to configure what the button does. At this point, at least with public link albums on iOS Safari, the button does nothing when pressed. When logged in it does work, but not clear where the photos go.

pulsejet:
> The upload button appears when sharing the album as public link

Ah that's a bug. Upload only works when logged in.

> When logged in it does work, but not clear where the photos go.

It prompts you where to store the files. The default is either the timeline path (the first one) or the current folder if you're in the folders view.

image

ChildLearningClub:
Okay, I see that now when I'm on the desktop Brave browser, but I'm given no such option when logged in on iOS Safari and uploading using that button.

ChildLearningClub:
edit wait... now it just popped up.

ChildLearningClub:
although in a future release the functionality to upload via public links would probably be wanted and would be per album. Those options could be put in under the "Share Album" section when sharing the album.

pulsejet:
> although in a future release the functionality to upload via public links would probably be wanted and would be per album

That'll be nice yeah. It's actually relatively straightforward for folder shares; albums are more work because of the need to specify where the files would go (and this would likely need support from Photos too)

PR #1076, opened 2024-03-19
Fix: correct group assigment for HW Acceleration

distante:
`/dev/dri/renderD128 does not belongs to the `video` group, but to the `render` group. 

Following the old steps will cause an `VA-API device (/dev/dri/renderD128) has incorrect permissions on the Memories config` error


pulsejet:
This is device/setup dependent (`video` is the correct group on my server). The `(may be different)` note is trying to highlight that the group may be different; this could be clarified better.

distante:
> This is device/setup dependent (`video` is the correct group on my server). The `(may be different)` note is trying to highlight that the group may be different; this could be clarified better.

I have added a more detailed text.

gymnae:
That is really distro dependent and should be checked by the user

distante:
Yes, the last change specifies how to check it.

PR #1086, opened 2024-03-23
Add Nextcloud cronjob to Dockerfile

rasmuslos:
Right now only the memories cronjob gets added to the cronfile, the Nextcloud background job should be added, too, so that the user does not have to rely on the other alternatives like Webcron.


pulsejet:
That's already added upstream https://github.com/nextcloud/docker/blob/6151f60208d08fe7dbd28b6b00a3bcd1e626e7ff/28/fpm/Dockerfile#L18

rasmuslos:
I did not know this, sorry

PR #1122, opened 2024-04-05
android: show local files on older devices

### LINKED ISSUE #1000: [Android] local photos do not work on older Android version
**Describe the bug**
I setup Memories for me and my parents.
We all use Android phones.
Everything works fine for me and my father.

But on my mothers phone local photos are only shown as placeholders.

All the photos from the server are ok.
But the files that are only local on the phone get no preview and are not shown wehen you click in the photo in the Memories app.

The file name is shown however.
So there is no permission issue.

Metadata (EXIF) is not extracted either, if I remember correctly.

**To Reproduce**

* Configure Memories via Web on the server
* install Memories app on Android phone
* setup app on phone
* see photo timeline with working and not working photos

**Screenshots**
image

**Platform:**

- OS: Android 10 (Samsung Galaxy S 9)
- Browser: Chrome (on WIndows 10)
- Memories Version: 6.1.5
- Nextcloud Version: 28.0.1
- PHP Version: 8.2.14

**Additional context**

From [Discord](https://discord.com/channels/1100157156742926406/1100158171621904567/1194062337573916742):

> pulsejet — 09.01.2024 18:58
> This could be because of the older Android version not supporting the preview API we're using, but it's a bug regardless. The blocker here is that someone needs to get a similar device / os and fix it.

According to https://www.appbrain.com/stats/top-android-sdk-versions this would affect ~1/3 of all Android devices if this bug affects all Android versions <11.

yamanq:
I found that the [previous PR](https://github.com/pulsejet/memories/pull/1118) was only one step to fixing the issue. Just like before, android:requestLegacyExternalStorage is needed for permission purposes, which fixes the EXIF data and full image on zoom. However, the thumbnail generation was still not working as the URI returned by ContentUris.withAppendedId was incorrect, so I used a different approach to get the correct URI.

I could not easily test on newer devices but on my Android 10, this fixes #1000.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -18,6 +18,7 @@
         android:supportsRtl="true"
         android:theme="@style/Theme.Memories"
         android:usesCleartextTraffic="true"
+        android:requestLegacyExternalStorage="true"
         tools:targetApi="31">
         <activity
             android:name=".MainActivity"
--- a/android/app/src/main/java/gallery/memories/service/ImageService.kt
+++ b/android/app/src/main/java/gallery/memories/service/ImageService.kt
@@ -7,6 +7,7 @@ import android.graphics.ImageDecoder
 import android.os.Build
 import android.provider.MediaStore
 import androidx.media3.common.util.UnstableApi
+import gallery.memories.mapper.SystemImage
 import java.io.ByteArrayOutputStream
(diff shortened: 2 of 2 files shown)


pulsejet:
Tested on 14, works well. Thank!

asandikci:
Is it possible to make it work in Android 9 also?

PR #1126, opened 2024-04-06
fix: Ensure read-write permissions for devcontainer

pktiuk:
It fixes my problem when I tried to develop something using devcontainer and GitHub Codespaces.

obraz


pulsejet:
Thanks!

pulsejet:
This doesn't seem to have fixed the issue and there doesn't seem to be any easy way around it. I've just added the fix to the docs https://github.com/pulsejet/memories/tree/master/.devcontainer

Issue #1111, opened 2024-04-02
Add an option to organize photos by year/month when moving photos


AiroPi:
**Is your feature request related to a problem? Please describe.**
When we work with a lot of photos, it is always better to have them organized by year/month or any date-related folder structure rather than all photos in a single folder. The mobile apps have some options to sort the photos by month during imports, but these apps (the iOS one at least) are very laggy, and crashes occur frequently. Also, when we import photos from a camera for example, we don't necessarily have any photo organization.

**Describe the solution you'd like**
This is why it would be awesome to have a little option to organize photos by date when we move some selected photos using Memories.

**Describe alternatives you've considered**
There are some software programs that are able to manage the file structure (Lightroom, Digicam...), these are big softwares with a lot of other functionalities, so it is a little bit "overkill" to use them to organize photos by date. Some scripts could be possible, they will mostly rely on webdav access to the file, which is quite slow to move and read a lot of different files. I didn't find any solution to organize the photos "from the inside" of Nextcloud.

Issue #595, opened 2023-04-20
Extend move to folder dialogue to allow copying as well


schlop:
**Describe the solution you'd like**
It would be great to have an option to not only move photos that are selected in the memories app to another folder, but also copy them to another folder. Would it be possible to add a copy button to the following dialogue?

<img width="596" alt="image" src="https://user-images.githubusercontent.com/6044565/233253117-da186189-560d-4f44-a657-a9e782788b80.png">

**Why would you want this?**
Copying photos from the memories timeline to another folder could be extremely useful to allow for the better sharing of photos with others. I currently backup my photos directly from my phone into memories. If I go on a trip with friends, it would be great to be able to copy all my photos that I took during the trip into a folder outside of the memories timeline folder. I could then share this folder with my friends, and would not have to worry about random photos appearing in my timeline, or photos being accidentally deleted by someone. Other use cases would be to copy/export photos out of memories for editing purposes, archival purposes, etc.

I can see that there is a risk that users might copy their files into another folder that is also displayed in the timeline, and then end up with a lot of duplicates. This could potentially be mitigated through a warning message. 

Generally, this might not be a high-priority feature but I think it would enable some advanced workflows!

akmail:
to myknowledge, it already supports copying: if I click on the move button, it asks for the destination and then I have both COPY and MOVE options. It looks like it just needs to be renamed to "copy or move". Am I right?

PR #1143, opened 2024-04-16
Add copy and "move and organise" features

AiroPi:
This implements features #595 and #1111. It's my first time working with TypeScript, PHP, and Vue, so this is still a draft. I'm aware there are many things that need refactoring.

This adds two new buttons to the "move" modal called “Move and Organize” and "Copy".

To implement these features, I needed to make a few changes:

- To accommodate adding multiple buttons, I refactored the `chooseNcFolder` function slightly. It now takes a `buttonFactory` instead of `type` as the third argument. `setType`was deprecated anyway in `getFilePickerBuilder`. The default value still behaves the same (providing the "Choose" option) so it doesn't break anything. Other types need to be defined manually.

> Another approach could have been to keep a memories-defined `type` enum, and have button implementations defined based on the type passed. However, determining which button was pressed would be trickier with multiple buttons. The current approach however needs more work in order to have different buttons than "Choose".

> I didn't implement the "Move to `x`" when a folder is selected, since with 3 buttons this can add a lot of text.

- To optimize the movement of files to different folders, I also created a `movePhotosByDate` function for `dav`. This function first creates the required directories, then moves the files.

> Initially, I considered a more general `movePhotosToDestinations` function that takes a `Map<Path, IPhotos[]>`, but this would require additional calls to create all the directories, check if they exist, and then create them. However, this approach could also allow for user-defined structures.

By the way, the new `copyPhotos` and `movePhotos` functions are very similar and could be merged.


AiroPi:
By the way, here is how it looks like:
image

pulsejet:
I intended to get to this before the 7.3 release but a hotfix came up that'll need an immediate release. Moving this ahead but 7.4 can come earlier.

pulsejet:
Happy new year 😄

shalak:
I'm trying to find some documentation, but with no luck. What does exactly "Move and organize" do? How does it differ from "Move"?

AiroPi:
> I'm trying to find some documentation, but with no luck. What does exactly "Move and organize" do? How does it differ from "Move"?

Some documentation could be welcomed but U don't really know where it could be. There is a little website for the documentation, maybe here.

To respond to your question, this feature is the implementation of #1111 which will not only move the photos, but also create a directory tree, sorting photos by year / months.
So in the memories perspective you won't notice the difference, but when you explore in folder view you will view the structure created.

shalak:
> Some documentation could be welcomed but I don't really know where it could be present

A tooltip on hover for the "Move and organise" button might be the most intuitive option. I'm not familiar with the codebase, but if the button factory supports a title field, that might be the simplest solution.

image


> To respond to your question, this feature is the implementation of https://github.com/pulsejet/memories/issues/1111 which will not only move the photos, but also create a directory tree, sorting photos by year / months.

Thanks — makes sense. Just to clarify: is the timestamp taken purely from EXIF metadata, or does it fall back to file modification times or other sources?

AiroPi:
> Thanks — makes sense. Just to clarify: is the timestamp taken purely from EXIF metadata, or does it fall back to file modification times or other sources?

It will respect the same rules than Memories use to sort the timeline, so yes, it fall back to file modification times etc...

Chikamatsu:
@AiroPi Thank you for this feature!

I am currently migrating my photo collection to Nextcloud and therefore have many photos that need to be properly organized. I upload the photos to an upload folder, select them all, and then apply Move and Organize to move them. Normally, I select and move about 200 to 300 photos at once.

But then, sometimes pictures are being moved to incorrect folders. For example, a picture from 2005 is moved to a folder for 2006, and with the months, there is, for instance, a folder 01 (January) that contains pictures from various months.

When i select the wrong pictures (only about 10) an use Move and Organise on this pictures again it seems to work correct.

Did anyone else notice this behavior?

edit: Now i can reproduce the behaviour and can open a issue: [https://github.com/pulsejet/memories/issues/1544](https://github.com/pulsejet/memories/issues/1544)

PR #1184, opened 2024-05-10
Update de.js

juewe:
Typo in prominent pos


pulsejet:
Thanks for your contribution. Nextcloud’s translation system is powered by Transifex. To start translating sign up and enter a group.

https://www.transifex.com/nextcloud/nextcloud/memories/

We cannot merge those changes directly because they would be reverted by the state of transifex on the next sync, so please contribute those to the translation system directly.

PR #1185, opened 2024-05-10
Android: Added onSaveInstanceState

tenJirka:
For some reason, I wasn´t able to find out why, on my phone (Pixel 7, android 14) is main activity always recreated when screen is rotated, and app always forget in which state it was. I was not able to replicate this behavior in android studio emulator, which is weird. It is really annoying so I decided  to add feature to restore MainActivity from previous state.

For the same reason no full screen setting (hide status bar) was called so I added it to onCreate function. All edits would be called once on devices without this problem, which should be fine, and this will work as back up for devices which do have same problem. 

Bonus is, that this will work every time, that for example split screen should now work better (App will not "restart").

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -24,7 +24,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:theme="@style/Theme.Memories.NoActionBar"
-            android:configChanges="orientation|screenSize|screenLayout">
+            android:configChanges="mcc|mnc|locale|touchscreen|keyboard|keyboardHidden|navigation|screenLayout|fontScale|uiMode|orientation|density|screenSize|smallestScreenSize">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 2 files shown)


tenJirka:
After adding screenLayout into manifest, app works properly even on my phone + it handles better split screen transformations.
Not sure why it was needed on my phone and not in the emulator.

Previous changes is not so much needed now, but in my opinion it is probably better to have in the app as backup if anything other will appear.

pulsejet:
Good catch. I'd remove the state saving if it's not required. Just saving the url is not sufficient state (that would require perfect deep linking, and there's other things that we rely on like browser history state). If it gets triggered it might give rise to more subtle bugs (having a more "noisy" issue is always better)

pulsejet:
Probably worth adding more props from here
https://developer.android.com/guide/topics/manifest/activity-element

We should be able to handle basically any change.

tenJirka:
I understand. I removed the state saving and added all the props from the link you suggested (hope I understood you correctly). I left the fullscreen setting in onCreate() to compensate for opening from landscape.

pulsejet:
Thanks @tenJirka

PR #1245, opened 2024-07-16
Display album metadata (location field, collaborators)

jovanbulck:
This is a draft PR that attempts to improve memories' albums view by displaying essential metadata.

## Added UI

First 2 commits attempt to display the "shared" status of an album. This makes it easier to see which albums are private or not and which users have access/contributed to a specific album.

1. show whether an album is shared in the albums overview listings (i.e., add a "shared" keyword, similar to how it's done for Google Photos and [nc-photos](https://gitlab.com/nkming2/nc-photos)).  (aside: not sure what is desired re translations, but I went for "Shared Album" for now as that is already in the translations currently; adding a more brief "Shared" translation keyword would probably be even more concise in the UI though..)
2. display the avatars of album collaborators above the album contents timeline (again similar to Google Photos)

Lastly:

3. display the "location" field already optionally provided by the user for every album (ie similar to how it's done already in Nextcloud Photos).

Point (3) above is especially lacking as the existing UI allows the user to record a "location" field but this is never displayed AFAIS..

## Screenshots

Screenshot from 2024-07-17 00-46-25
Screenshot from 2024-07-17 00-50-15

## Remaining Issues

This is my first attempt at diving into Nextcloud/PHP/Vue/etc so feedback is greatly appreciated!

~A remaining problem with the current code is that points (2-3) above require the metadata to be in `$route.params`, but it seems like one can only put it there when navigating to the album page via the albums cluster listing (cf. my hack in `getClusterLinkTarget` in `clusters.ts`). Thus, the added metadata (2nd screenshot) is not displayed when navigating to the album directly via a link (or reloading the page). This is clearly undesirable, but I couldn't figure out how to get around that with my limited understanding of Vue.~

~@pulsejet or others: could you give some directions on what is the proper way to get to the `IAlbum.location` and other properties from within the DynamicTopMatter? I'd be happy to work further on this PR to refine the functionality.~

### DIFF
--- a/lib/Db/AlbumsQuery.php
+++ b/lib/Db/AlbumsQuery.php
@@ -88,12 +88,20 @@ class AlbumsQuery
         // FETCH all albums
         $albums = $query->executeQuery()->fetchAll();
 
+        // Additionally fetch all shared album ids
+        $queryShared = $this->connection->getQueryBuilder();
+        $queryShared->select('album_id')->from($this->collaboratorsTable());
+        $shared_ids = array_map(function($row) {
+            return (int)$row['album_id'];
+        }, $queryShared->executeQuery()->fetchAll());
+
         // Post process
         foreach ($albums as &$row) {
             $row['cluster_id'] = $row['user'].'/'.$row['name'];
             $row['album_id'] = (int) $row['album_id'];
             $row['created'] = (int) $row['created'];
             $row['last_added_photo'] = (int) $row['last_added_photo'];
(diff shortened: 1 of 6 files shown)


pulsejet:
Thanks a lot!

Made a few changes for the merge -- mainly we don't need the separate query to get the "shared" flag, this can be done in the same subquery, and fixed links in the shared view.

jovanbulck:
Awesome, thanks for making the changes and merging!

Issue #787, opened 2023-08-23
Add `Shared with you` menu item


meichthys:
**Is your feature request related to a problem? Please describe.**
This is to match default Photos app functionality.

**Describe the solution you'd like**
Add a `Shared with You` menu to the main menu:
![image](https://github.com/pulsejet/memories/assets/10717998/068f82f3-206d-4df3-a4e3-a303eccdb56a)

**Additional context**
This is available in the default Photos app

PR #1266, opened 2024-08-11
feat: Add username label on shared photos

### LINKED ISSUE #955: Add username label in shared album for added photos
When I'm using shared albums I'm missing information who is owner of the photo, or if preffered who added photo to album. Simply I don't like to use somebody else's photo as my own and then provoke any misunderstandings between users sharing photos in one album.

I would like to have possibility of turning on photos labelling on preview and / or sorting by user / owner.

Now I need to enter "info" and the only thing I can do is to check camera type, sometimes I can find out if it is my photo or I recieved it from somebody, but it is quite frustrating and not effective way.

Similar feature exist in Google Photos, please find photo below:

Screenshot_20231208-123226617 (1)

jovanbulck:
Fixes #955

I find this feature really lacking when sharing large albums and/or timelines with multiple people, so I had a shot at implementing this the other day. Currently looks as follows:

Screenshot from 2024-08-10 21-26-00

The nextcloud username is shown in the bottom left of the picture frame for all pictures that are not owned by the logged-in user, both in the Timeline and Albums view.

I had to extend the database scheme with an additional "owner" column that stores the owner UUID for every memories-indexed photo. This would probably require a `lib/Migration/` handler to amend existing rows; not sure how this is normally handled -- @pulsejet ?

Possible extensions include:

* displaying full user info (ie not just the nextcloud username) in the "metadata" pane
* filtered view for a dedicated "Shared with you" menu item cf #787 -- (one of the only features that Photos has that Memories doesn't..)

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -56,6 +56,8 @@ return [
         ['name' => 'Map#clusters', 'url' => '/api/map/clusters', 'verb' => 'GET'],
         ['name' => 'Map#init', 'url' => '/api/map/init', 'verb' => 'GET'],
 
+        ['name' => 'Uid#name', 'url' => '/api/uid/name/{id}', 'verb' => 'GET'],
+
         ['name' => 'Archive#archive', 'url' => '/api/archive/{id}', 'verb' => 'PATCH'],
 
         ['name' => 'Image#preview', 'url' => '/api/image/preview/{id}', 'verb' => 'GET'],
--- a/lib/Controller/UidController.php
+++ b/lib/Controller/UidController.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2022 Varun Patil <radialapps@gmail.com>
+ * @author Varun Patil <radialapps@gmail.com>
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\Memories\Controller;
+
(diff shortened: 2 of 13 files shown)


pulsejet:
This is very nice. Populating the initial data should be easy, just need a loop to copy over during a migration. I'm not aware of any cases where file ownership can transfer so that should also be ok.

The only thing remaining is to make sure we show the display name of the user (not the username). This can be done with a client side cache (per-session)

Btw it might be a while before I can get to merging this. I'm not allowed to work on this project for a bit more time.

jovanbulck:
Great, thanks! No worries about having this in the queue for a while before you can get to it :)

When you get to it, feel free to let me know if there is something I can help with from my side to get this merged

jovanbulck:
> This is very nice. Populating the initial data should be easy, just need a loop to copy over during a migration. I'm not aware of any cases where file ownership can transfer so that should also be ok.

Update (force-pushed in the above commit): I added  a simple Migration handler to do this. It gets the UID from the `oc_storages` table, based on the storage ID found in the `oc_filecache` table for the respective file referenced in `oc_memories`.

I tested this on my end with Postgres and it ran very fast on a large DB. However, not sure about the exact syntax for SQLite and MySQL. Also not sure about edge cases like External Storage in Nextcloud (right now the queries rely on the `home::uid` syntax in `oc_storages`.  FWIW, I made the new UID column notnulll=false in case some of these edge cases wouldn't work.

> The only thing remaining is to make sure we show the display name of the user (not the username). This can be done with a client side cache (per-session)

This would still have to be implemented. Not sure how best to go about it, but maybe an initial query that loads all unique `uid`s from `oc_memories` and then translates the displayname from `oc_users` and send it back to the client could work(?)

jovanbulck:
>  The only thing remaining is to make sure we show the display name of the user (not the username). This can be done with a client side cache (per-session)

This is now also implemented (as per the force-pushed latest commit above) using a new remote `UidController` API for the translation and local `utils/cache` to speed this up for repeated translations. This was new to me, so please double check :)

@pulsejet Marking this as "ready for review" as this PR should now be feature-complete and works stable on my end (both on a postgresql and a mysql setup).

pulsejet:
A slightly different implementation is merged (you're co-author!) in https://github.com/pulsejet/memories/commit/eb7fa255b8fb6eb9a774d1b700ff1f5b05929841

Storing the uid in the memories table (and especially the migration) seems somewhat risky. Instead we can directly query the storag table to extract the UID during the "day" query phase.

jovanbulck:
Awesome, thanks for merging this upstream! Avoiding the UID storage via `storage_id` is indeed neater :)

endoflevelboss:
Is there a way to disable the name display? The test field may be unwanted for esthetic reasons in the timeline view. In timelines shared by public links, the display also leaks internal information by disclosing the internal user's ful name.

PR #1270, opened 2024-08-15
go-vod: update hw transcoding advice for wsl

MondoGao:
On Windows, users don't need to install the container-tookit to enable gpu support in container. Add a verified tip for this senerio.


pulsejet:
Thanks!

PR #1311, opened 2024-10-11
Devcontainer

jovanbulck:
Some small changes to easy development:

* enabled [Adminer](https://www.adminer.org/) on port 8080 for easy graphical database inspection/management
* clarifies how to make sure updated app is loaded and not older cached version


pulsejet:
Nice, thank you!

Issue #285, opened 2022-12-12
TODO: Face Recognition Integration.


matiasdelellis:
Since Memories v4.9.2 we have an integration with the [Face Recognition](https://github.com/matiasdelellis/facerecognition/) as alpha. Now just open this issue to track pending tasks, and handle any user expectations.

When I proposed this integration, we try to share as much code and dialogs as possible, in order to have a consistent workflow in both applications. There were some features that both apps work differently that we couldn't make them compatible for this stage and we hope to add them soon.

**TODO:**
- [ ] Fix "Merge with different Person". Today it shows the dialog correctly, but fails miserably. 😅 In the meantime, I recommend you assign name directly :wink:. The fix is easy, but requires a "small" rewrite of the dialog.
- [ ] Implement "Move to another person".
- [ ] Implement "Remove from person". Issue #283 

**Required api:**
* "Merge with different Person"
  - PUT: /index.php/apps/facerecognition/api/2.0/cluster/317878?name=Name
  - PUT: /index.php/apps/facerecognition/api/2.0/person/Penny?name=Not+Penny
  - The service already exists `await dav.renamePeopleFaceRecognition(oldName/clusterId, nameOfOtherPerson);` but we have to adapt the dialog.

* "Move to another person":
  - PUT: /index.php/apps/facerecognition/api/2.0/face/{faceId}/detach?name=New+Name
  - It requires the id of the face, so you know exactly who you are moving. At the moment we do not have this info in the queries.

* "Remove from person"
  - PUT: /index.php/apps/facerecognition/api/2.0/face/{faceId}/detach
  - It requires the id of the face, so you know exactly who you are moving. At the moment we do not have this info in the queries.

p.s: I'm speaking as Face Recognition developer, and just trying to help here. 😉

qudiqudi:
Really looking forward to see this working! 😃

matiasdelellis:
Hi @qudiqudi 
Thanks for reporting it... The report is very similar to #35

It seems to be a limitation of PostgreSQL.. Well, far from being a limitation, just mysql is more flexible, and same query seems to be confusing for PostgreSQL and rejects it. 😅 

I can't reproduce it, but please try this. It is the only inconsistency that I see with the naked eye.
```
diff --git a/lib/Db/TimelineQueryPeopleFaceRecognition.php b/lib/Db/TimelineQueryPeopleFaceRecognition.php
index 2676b87..086a801 100644
[...]

PR #1339, opened 2024-11-06
Implement merge person/clusters on Face recognition backend. Part of …

matiasdelellis:
…issue #285

Just try to make the patch as simple as possible without changing a single line of behavior in recognize.

The warning that you can only merge with named people is important, and while I personally don't understand why we would merge unnamed people, it is a feature that the recognize backend supports.

### DIFF
--- a/src/components/modal/FaceMergeModal.vue
+++ b/src/components/modal/FaceMergeModal.vue
@@ -99,6 +99,20 @@ export default defineComponent({
         return;
       }
 
+      if (this.routeIsFaceRecognition) {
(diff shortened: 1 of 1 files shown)


Rudd-O:
Can we please have a release with this fix?  Thanks.

Rudd-O:
I can confirm the fix worked.  I produced a prerelease, untarred it into my apps folder, and BAM, it worked.

Here is a link to the file: https://cloud.rudd-o.com/s/seiEZdAyZnkJzm7

Issue #1036, opened 2024-02-08
Integration with Photosphere viewer


SayantanRC:
**Is your feature request related to a problem? Please describe.**
Memories cannot display photospheres.

**Describe the solution you'd like**
Integrate with [Photosphere viewer](https://github.com/nextcloud/files_photospheres). Nextcloud Photos switch to Photosphere viewer automatically when opening a 360 degree photo.

**Describe alternatives you've considered**
None because Photosphere viewer is the only maintained app for this purpose.

**Additional context**
N/A

major-mayer:
Ohh great, I didn't even know that such an app exist in the Nextcloud ecosystem.
But now that i know, I would like to see such a feature in Memories as well :heart:

Issue #931, opened 2023-11-20
Photo Sphere Viewer


rondadon:
**Is your feature request related to a problem? Please describe.**
No, there is no Problem related to the feature request. 

**Describe the solution you'd like**
It would be nice when opening a Spheric Photo (Photo Sphere) it would be shown as a sphere where you can drag the sphere around.

**Describe alternatives you've considered**

**Additional context**
There is a php Library which processes such Photo Spheres. Maybe this could be implemented?
**Website:** https://photo-sphere-viewer.js.org/
**Demo:** https://photo-sphere-viewer.js.org/demos/
**Github-Repo:** https://github.com/mistic100/Photo-Sphere-Viewer

There is even a Nextcloud Photo Sphere Plugin (based on previously mentioned php library) but only works with the stock photo feature of nextcloud:
https://apps.nextcloud.com/apps/files_photospheres

Thank you and have a nice day! :)

Issue #676, opened 2023-06-01
Panorama Sidebar Menu


meichthys:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to filter photos by panorama images.

**Describe the solution you'd like**
It would be nice to be able to have a new panorama view that would show all panoramic photos in timeline view. For example,  iCloud Photos has this view:
![image](https://github.com/pulsejet/memories/assets/10717998/c8b64004-8706-447a-b97a-b3611b7efaab)

**Describe alternatives you've considered**
Scanning the timeline isn't ideal.

**Additional context**
I suppose we would need to base this off the photo resolution. Perhaps a photo more than 3:1 would be considered a panorama. [Wikipedia defines](https://en.wikipedia.org/wiki/Panoramic_photography) it as 2:1 but some phones approach that with regular photos.

PR #1357, opened 2024-11-28
feat: Add separate categories for "live photos" and "panoramas"  and support for 360 PhotoSphere viewer

jovanbulck:
Add the following features:

* separate API and category in the "explore" menu for live photos and panoramas cf #676
    - includes a heuristic to identify "panoramas" based on >= 3:1 aspect ratio
* full integration with [Photo Sphere Viewer JavaScript library](https://photo-sphere-viewer.js.org/) cf #1036 and #931  
    - this can be opened from a new "open in PhotoSphere" button in the image viewer
    - Tested this on mobile and desktop on my end.

Possible further improvements:
* only show the  "open in PhotoSphere" button for pictures that we labeled as "panoramas"?
* parse the XMP metadata to discover actual photospheres, even if they are < 3:1? This is what the `files/photosphere` NC app [seems to do](https://github.com/nextcloud/files_photospheres/blob/master/lib/Service/Helper/XmpDataReader.php). In principle, this could be done server side and cached in the `oc_memories` table, but not all panorama pictures include XMP data afaik.

image
image

### DIFF
--- a/lib/Db/TimelineQueryFilters.php
+++ b/lib/Db/TimelineQueryFilters.php
@@ -9,6 +9,11 @@ use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\DB\QueryBuilder\IQueryFunction;
 use OCP\ITags;
 
+// Wikipedia defines a panoramic image as having an aspect ratio of at least 2:1,
+// but some phones approach this with regular photos. Hence, we conservatively set
+// the threshold to 3:1 for true panoramas.
+const PANOROMA_ASPECT_RATIO = 3;
+
 trait TimelineQueryFilters
 {
     public function transformFavoriteFilter(IQueryBuilder &$query, bool $aggregate): void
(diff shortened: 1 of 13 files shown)


lib/Db/TimelineQueryFilters.php
```diff
@@ -37,6 +42,16 @@ public function transformVideoFilter(IQueryBuilder &$query, bool $aggregate): vo
...
     }
 
+    public function transformLivePhotoFilter(IQueryBuilder &$query, bool $aggregate): void
+    {
+        $query->andWhere($query->expr()->neq('m.liveid', $query->expr()->literal('')));
+    }
+
+    public function transformPanoFilter(IQueryBuilder &$query, bool $aggregate): void
+    {
+        $query->andWhere('m.w >= '.PANOROMA_ASPECT_RATIO.' * m.h');
```
pulsejet, line 52: Both of these will scan the entire table - these need the right indexes.

pulsejet: AFAIK this expression can't be optimized with index too, probably needs a separate boolean field.

jovanbulck: thanks for the feedback, makes sense! If I find some time later, I'll try to add a `pano` boolean field in the `oc_memories` table

baka0815:
@jovanbulck are you still working on this? Is there anything you need help with?
I would love to see this feature in Memories.

jovanbulck:
> @jovanbulck are you still working on this? Is there anything you need help with? I would love to see this feature in Memories.

I'm sorry, but I'm not working on this anymore (switched to a different photo management system). Though you're very welcome to fork my fork and continue development as you see fit!

baka0815:
I got it working on my side and might take over your work.

TychoGold:
@baka0815 @pulsejet 

Hi everybody, is this already working? It would be *really* great to have this in memories as this is one of the only features that "Photos" is giving. Everything else works better in memories. (talking about 360°)

PR #1435, opened 2025-04-03
chore: also test with NC v31 and php-lint with 8.3

Croydon:


Croydon:
I have splitted out test coverage for NC 31: #1436

arnowelzel:
The standard for functions, classes and methods is to put the opening braces (curly brackets) in separate lines and not next to it. Also see the examples at https://www.php-fig.org/psr/psr-1/ and https://www.php-fig.org/psr/psr-12/.

Example - this is common practice:

```
class MyClass
{
    function myMethod()
    {
        $x = doSomething();
        if ($x < 5) {
            doSomethingElse();
        }
    }
}
```

But not this, because it is then harder to recognize the start of a method compared to the rest of the code:

```
class MyClass {
    function myMethod() {
        $x = doSomething();
        if ($x < 5) {
            doSomethingElse();
        }
    }
}
```

Croydon:
Changes in this PR are either merged at this point or are no longer needed.

Closing in favor of #1502

PR #1441, opened 2025-04-13
[Bug] Takeout: Invalid JSON name

Javinator9889:
The Google Takeout feature has changed. The downloaded tarball will contain the JSON metadata but with a different name, including - fully or partially - the string "supplemental-metadata" before the `.json` terminator.

This commit modifies the logic to look for a file matching a RegExp (https://regex101.com/r/rXUYj4/1) rather than just appending a `.json` suffix.

### DIFF
--- a/lib/Command/MigrateGoogleTakeout.php
+++ b/lib/Command/MigrateGoogleTakeout.php
@@ -167,7 +167,7 @@ class MigrateGoogleTakeout extends Command
                 $this->migrateFolder($node);
             } elseif ($node instanceof File) {
                 try {
-                    $this->migrateFile($node);
+                    $this->migrateFile($node, $nodes);
                 } catch (\Exception $e) {
                     $this->output->writeln("<error>Error migrating file {$node->getPath()}: {$e->getMessage()}</error>");
                 }
@@ -175,7 +175,7 @@ class MigrateGoogleTakeout extends Command
         }
     }
 
-    protected function migrateFile(File $file): void
+    protected function migrateFile(File $file, array $nodes): void
     {
         // Check if this is a supported file
         if (!\in_array($file->getMimeType(), $this->mimeTypes, true)) {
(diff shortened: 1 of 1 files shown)


Javinator9889:
A quick note: The RegExp looks for any character in `supplemental-metadata.` - including the dot. This is **not robust** and can match any other word containing any of those characters. However, it's simple enough to just work.

I did my tests just using `[a-zA-Z0-9_]*`, which is even wider but simpler.

lib/Command/MigrateGoogleTakeout.php
```diff
@@ -190,7 +190,29 @@ protected function migrateFile(File $file): void
         $jsonFile = null;
 
         try {
-            $jsonPath = $path.'.json';
+            /* the JSON file may contain the "supplemental-metadata" string, fully or partially */
+            $nodes = $this->rootFolder->getDirectoryListing();
```
pulsejet, line 194: I'm somewhat concerned about this since it will enlist the directory for each file - if you've large directories this is likely too slow with O(n^2)

pulsejet (commented):
Thanks for your contribution. Trust Google to screw things up :(

It might be a while before I get a chance to test this out, but I'll try to get to it asap.

lib/Command/MigrateGoogleTakeout.php
Javinator9889: You're totally right, I've addressed your comments

Javinator9889:
> Thanks for your contribution. Trust Google to screw things up :(
> 
> It might be a while before I get a chance to test this out, but I'll try to get to it asap.

No worries! Thanks a ton for your time

I've been playing around and trying to develop a unit test for the new behavior, but as the exposed API is always behind a command, it's hard to do so. It'd end-up being a functional, black box test, which I don't think it's quite useful at this stage

Javinator9889:
For the moment, I've addressed the invalid name problem by using shell magic. It should just work until this PR gets merged :)

```shell
# This script needs Bash >4

# A small function to remove the "supplemental-metadata" string
rename-json() {
    local orig=$1
    local dst=$(sed -E "s/(.+)\.[supplementalmetadata-]+\.json/\1.json/g" <<< "$orig")
    mv -v "$orig" "$dst"
}

readarray -d '' json_files < <(find . -iname '*.json' -print0)
for file in "${json_files[@]}"; do
    rename-json "$file"
done
```

I left that running in the background, got all files renamed, and then just simply ran `php occ memories:migrate-google-takeout`. It worked like a charm, having all my photos with the correct EXIF extracted from JSON files

PR #1478, opened 2025-07-14
Use system default for timezone

### LINKED ISSUE #1240: Wrong timestamp pictures. Off by 2 hours
### Describe the bug

When auto uploading pictures the picture has the wrong timestamp in memories. (Its 2 hours off)

The server timezone is correct as is also the nextcloud timezone.

### Steps To Reproduce

Always

### Platform

```markdown
- OS:debian 12
- Browser: native android app
- Memories Version:latest
- Nextcloud Version:latest
- PHP Version:8.2
```

### Screenshots

_No response_

### Additional context

- Any errors in the JS console?
- Any errors in the Nextcloud server logs?

yamanq:
The TZ environment variable is not set on many machines and date_default_timezone_get only works if the timezone has been set elsewhere. Instead, this approach relies on the Nextcloud server configuration for the timezone.

Closes #1240

See: https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#default-timezone

### DIFF
--- a/lib/Db/TimelineWrite.php
+++ b/lib/Db/TimelineWrite.php
@@ -9,6 +9,7 @@ use OCA\Memories\Service\Index;
 use OCA\Memories\Util;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\File;
+use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\Lock\ILockingProvider;
 use Psr\Log\LoggerInterface;
@@ -28,6 +29,7 @@ class TimelineWrite
         protected LivePhoto $livePhoto,
         protected ILockingProvider $lockingProvider,
         protected LoggerInterface $logger,
+        protected IConfig $config,
     ) {}
 
     /**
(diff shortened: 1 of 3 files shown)


lib/Db/TimelineWrite.php
```diff
@@ -121,7 +123,7 @@ public function processFile(
         [$lat, $lon, $mapCluster] = $this->processExifLocation($fileId, $exif, $prevRow);
 
         // Get date parameters (after setting timezone offset)
-        $dateTaken = Exif::getDateTaken($file, $exif);
+        $dateTaken = Exif::getDateTaken($file, $exif, $this->config->getSystemValue('default_timezone', 'UTC'));
```
pulsejet, line 126: This will always override the `getenv` fallback

lib/Exif.php
```diff
@@ -248,7 +248,7 @@ public static function getDateTaken(File $file, array $exif): \DateTime
         $dt = new \DateTime('@'.$file->getMtime());
 
         // Set timezone to system timezone
-        $tz = getenv('TZ') ?: date_default_timezone_get();
+        $tz = $timezone ?: getenv('TZ') ?: date_default_timezone_get();
```
pulsejet, line 251: Preferable to use `OCA\Memories\Settings\SystemConfig`

pulsejet (approved):
Thanks, I'll take care of the above comments before merge

PR #1506, opened 2025-08-03
Add the capability to multi-select from within the Viewer using UI or keyboard shortcuts.

mm-zacharydavison:
A feature I find essential in photo management software is the ability to quickly flick through a large set of photos and 'audit' them before I add them to my library permanently.

The common workflow in tools like Lightroom is to use keyboard shortcuts while fullscreened to multi-select a group that you can then do whatever you like with (e.g. delete).

I found this cumbersome in Memories, so added the functionality.

image
image

- There is an empty circle / checkmark in the action bar of the `Viewer` that indicates current selection status.
- You can press `' '` (space bar) to select/de-select current image.
- When you return to the timeline, your images will be selected.

It's my first time contributing to this repo, so please let me know:

1. If the approach makes sense (especially for the `selected` fragment handling).
2. If I should add any test cases (I saw no existing tests for the viewer).

### DIFF
--- a/src/components/SelectionManager.vue
+++ b/src/components/SelectionManager.vue
@@ -191,9 +191,14 @@ export default defineComponent({
     touchScrollDelta: 0,
     touchMoveSelFrame: 0,
     multiSelectDelta: null as 1 | -1 | null,
+    shouldPersistSelections: false,
   }),
 
   mounted() {
+    _m.selectionManager = {
+      selectPhoto: this.selectPhoto.bind(this) as typeof this.selectPhoto,
+    };
+
     // Make default actions
     this.defaultActions = [
       {
@@ -643,10 +648,12 @@ export default defineComponent({
         photo.flag |= this.c.FLAG_SELECTED;
         this.selection.addBy(photo);
         this.selectionChanged();
+        console.log('added photo', this.selection);
       } else {
         photo.flag &= ~this.c.FLAG_SELECTED;
         this.selection.deleteBy(photo);
         this.selectionChanged();
+        console.log('removed photo', this.selection);
       }
 
       if (!noUpdate) {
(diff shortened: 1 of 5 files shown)


mm-zacharydavison:
There's some more QoL keyboard shortcut things I'd like to add to this (e.g. having the delete shortcut delete your current selection, not just current photo) but I'll wait for feedback on this PR first.

pulsejet:
Seems good to me in general. A couple things that'll need to change

1. I'm trying to move away from globals, it's rather unfortunate that we have any at all. Ideally we need to be able to support multiple instances of Timeline on a single page (this is needed for some features like add to album). I've been putting this off to after the Vue 3 migration; there are new utils that'll make this easier - regardless we need to avoid adding another global. Probably makes more sense to use the event bus here (which can later be replaced by a viewer generated event)
2. The UI will need to be hidden since this is not a very commonly used feature. Probably a shortcut to trigger it and then it'll become visible should be fine. We need a page or popup to show all the hidden shortcuts anyway.

mm-zacharydavison:
> Seems good to me in general. A couple things that'll need to change
> 
>     1. I'm trying to move away from globals, it's rather unfortunate that we have any at all. Ideally we need to be able to support multiple instances of Timeline on a single page (this is needed for some features like add to album). I've been putting this off to after the Vue 3 migration; there are new utils that'll make this easier - regardless we need to avoid adding another global. Probably makes more sense to use the event bus here (which can later be replaced by a viewer generated event)
> 
>     2. The UI will need to be hidden since this is not a very commonly used feature. Probably a shortcut to trigger it and then it'll become visible should be fine. We need a page or popup to show all the hidden shortcuts anyway.

Sure i can implement #1, I was originally going to not use globals but followed your lead with the rest of the codebase.
For #2, some options, i'm happy to implement whichever you prefer:

1. app setting to enable multi-selection UI
2. once the first photo is selected with keyboard shortcut, start showing the multi-select checkbox in the action bar
3. another shortcut required to enable

personally I'd rather have #1 if we have to hide the UI by default. The UX for #2 and #3 aren't great.
if we go with #1, if people like it in future you could enable it by default

I wanted to go ahead and make other actions compatible with multi-select, if you're fine with that existing I'm happy to implement it (e.g. delete, share, etc would all operate on your multi-selected photos if you have an active multi-select).
share might be a bit of a different flow, and i dont really need it, i just want delete (which is already supported by just exiting the viewer and deleting from the Timeline anyway, so my use-case is covered already)

mm-zacharydavison:
@pulsejet any movement on this? I'm happily using my own fork for now, so not really in a rush.

PR #1523, opened 2025-08-31
Add upload button in public shares

gouverneurp:
**Feature: Anonymous uploads for public folder shares**

This pull request adds the ability for anonymous users to upload files to public folder shares.

The upload button will only appear on a public share link if it has been created with "Can edit" permissions, ensuring that folder owners retain full control. When a user uploads files, a progress bar is displayed. After the upload is successfully completed, the page reloads to show the newly added files.

Below are some screenshots of the process for uploading to a public share:

1. Visible upload button in a public share. No user is logged in. Screenshot from 2025-08-31 19-43-09
2. Menu when the upload button is pressed. Screenshot from 2025-08-31 19-30-33
3. Progress bar for the upload. Screenshot from 2025-08-31 19-36-55
4. New uploaded image is available in the folder. Screenshot from 2025-08-31 19-30-51

### DIFF
--- a/src/components/top-matter/FolderTopMatter.vue
+++ b/src/components/top-matter/FolderTopMatter.vue
@@ -32,6 +32,17 @@
           <template #icon> <UploadIcon :size="20" /> </template>
         </NcActionButton>
 
+        <!-- NEW PUBLIC UPLOAD BUTTON -->
+        <NcActionButton
+          v-if="routeIsPublic && allowUpload"
+          :aria-label="t('memories', 'Upload files')"
+          :disabled="isUploading"
+          @click="triggerFileUpload"
+        >
+          {{ isUploading ? t('memories', 'Uploading...') : t('memories', 'Upload files') }}
+          <template #icon> <UploadIcon :size="20" /> </template>
+        </NcActionButton>
+
         <NcActionButton @click="toggleRecursive" close-after-click>
           {{ recursive ? t('memories', 'Folder view') : t('memories', 'Timeline view') }}
           <template #icon>
@@ -41,11 +52,22 @@
         </NcActionButton>
       </NcActions>
     </div>
+
+    <!-- Hidden file input that we trigger programmatically -->
+    <input ref="fileInput" type="file" style="display: none" multiple @change="handleFileSelection" />
+
+    <!-- NEW UPLOAD PROGRESS DISPLAY -->
+    <div v-if="isUploading" class="upload-progress">
+      <progress :value="uploadProgress" max="100"></progress>
+      <span>{{ uploadStatus }}</span>
+    </div>
   </div>
 </template>
 
 <script lang="ts">
 import { defineComponent } from 'vue';
+import { generateUrl } from '@nextcloud/router';
+import { t } from '@nextcloud/l10n';
 
 import UserConfig from '@mixins/UserConfig';
 
@@ -80,6 +102,16 @@ export default defineComponent({
 
   mixins: [UserConfig],
 
+  data() {
+    return {
+      isUploading: false,
+      uploadProgress: 0,
+      uploadStatus: '',
+      uploadCount: 0,
+      uploadFailures: 0,
+    };
+  },
+
   computed: {
     list(): {
       text: string;
@@ -110,6 +142,10 @@ export default defineComponent({
     isNative(): boolean {
       return nativex.has();
     },
+    
+    allowUpload(): boolean {
+      return this.initstate.allow_upload === true;
+    },
   },
 
   methods: {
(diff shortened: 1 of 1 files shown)


major-mayer:
Great news, thanks for your effort.
I am happy testing this once it's in a testable shape 👌

Rdeisenroth:
Finally, my most anticipated memories feature is being worked on. This will make it a good alternative to Immich and Google photos for shared Family Event, Holiday trips and so much more.

A few more ideas that might make this feature even better (from my limited understanding not yet implemented, sorry if they are):
- Option to specify where uploaded files for the Album will be stored (ideally link album directly to a folder and sync both ways, but one way would already be really nice)
- Option to create two (or more) links - one for just viewing the album, and one that can edit. I don't know if it is feasible, but ideally this could be tied to nextclouds existing sharing feature, allowing password protection, expiration dates, etc.
- Option for a Quota/File type filters. Sometimes you don't want someone uploading 10GB of videos and 500 Raw photos

Thank you for your work. I am happy to test it, as soon as you deem it ready

gouverneurp:
To be honest, this PR is still a work in progress. I think I published it too early, and it will take a while longer until everything is ready.
@Rdeisenroth, thanks for the feedback. For now, I don't plan to implement all of these ideas, just the simple file upload. But these ideas can then be easily integrated.

gouverneurp:
@Rdeisenroth, regarding your comment:
You can create different links for sharing. Uploading to public shares by users without an account is only possible in those marked as ‘Editable’. 
Screenshot from 2025-09-07 17-38-13

gouverneurp:
@major-mayer, the first version is ready and working on my end.
Would you still like to test it?
Thanks in advance!

major-mayer:
Yes very much. I will try to get it done in the next few days.
Any special advice how to test it? I guess I will have to clone your branch, build it and deploy the package to my Webserver, right?

gouverneurp:
@major-mayer Yes, that would be a possibility. I personally have used the Nextcloud development environment (https://github.com/juliusknorr/nextcloud-docker-dev) and don't know if I would currently recommend using it in a nextcloud instance that is used productively.

@pulsejet, What do you think? How do you normally test?

major-mayer:
For the start, I installed it on my local development setup (Nextcloud 31).

A few quick notes:
- In general the upload feature works, but only when you have increased PHP's `post_max_size` and `upload_max_filesize` settings to a value that it can handle the uploaded file sizes
  -  otherwise the request will just be aborted with a 400 error and the not very helpful error message "No file uploaded or upload error" in the console logs
  - Multi-part upload would be quite important for larger videos, I guess
- The toast notification when something goes wrong disappear very quickly
- The upload progress bar could need some padding on the right side 
- For some reason, the page reloads once the upload is finished in my case
  - It would be much nicer, if just the file list gets updated 
- During link creation, it would be great if one could switch the created link to editable already in this window:
  - It's not very intuitive for users to do this in a different part of the app
  - As a minimum the user should be at least informed that he can make the link editable, by switching the flag in the sidebar
grafik

gouverneurp:
@major-mayer, thanks for testing. Here are my answers:
1. The post_max_size and upload_max_filesize parameters are settings for the Nextcloud instance. In my opinion, everything is working as intended here.
2. Do you mean the upload status in the upper right corner?
3. I tried to improve it: Screenshot from 2025-09-09 12-27-50

4. I updated it so that the page is not completely reloaded, but only the images are updated
5. I have not changed the “link creation” and do not currently plan to do so.

gouverneurp:
Quick update:
I've updated the design of the progress bar and tested it on Ubuntu with Firefox and Android with Chrome.

**Ubuntu**:
demo_ubuntu

**Android**:
demo_android

major-mayer:
> I've updated the design of the progress bar and tested it on Ubuntu with Firefox and Android with Chrome.

Looks better, I agree. I would suggest to align the icons vertically centered whenever the progress bar is not visible. 

>     1. The post_max_size and upload_max_filesize parameters are settings for the Nextcloud instance. In my opinion, everything is working as intended here.

I know that those are PHP variables, that apply to the whole web server.
My point is that I think it's not feasible, and for some managed hosting instances even possible, to increase these variables to a value that can cover all image and especially video sizes you would encounter.
That's why I suggested using the multi-part upload (or chunked upload in Nextcloud terms) APIs that Nextcloud provides, which effectively handles the problem: https://docs.nextcloud.com/server/latest/developer_manual/client_apis/WebDAV/chunking.html
This is important for video uploads where file sizes can reach easily multiple gigabytes, which the PHP might not be able to handle using a single request depending on the server specs. 


>     4. I updated it so that the page is not completely reloaded, but only the images are updated
Looking better now, but still not as smooth as with the existing, authenticated upload functionality in Memories.
It seems like your PR reloads all images, while the existing upload adds the missing files to the folder view, resulting in less content flashes.

>     5. I have not changed the “link creation” and do not currently plan to do so.
Fair enough, but I think that's a pity and the PR would benefit greatly from it.


In general, I wonder why the seemingly existing logic and UI (upload button, progress bar) for uploading files (but in an authenticated state) in Memories cannot be re-used.
Of course, you wouldn't want to give the user to select the upload path in a public share, but apart from that... Might make things easier.

If necessary, it seems to make use of chunked uploading as well: 
grafik

gouverneurp:
Hey @major-mayer, 
Now the PR uses the Nextcloud uploader, so hopefully it will work better now! The automatic refresh of the page after uploading should also run more smoothly.
Best

major-mayer:
Hi @gouverneurp,
I just tested your latest changes, and it's really a big improvement in comparison to the last versions.
I am now able to upload videos with hundreds of megabytes, even tho the PHP settings allow only much smaller upload sizes.
The refresh process is also as expected.

The only remaining thing that I noticed is that when you change the link type to "edit" mode, it allows you to upload new files - as expected - but it doesn't allow you to delete existing files using the external share.
I could imagine that users of the external file share regularly upload images accidentally and need to delete them afterward.
IMO this should only be the case, if the user creates a link using custom share options, that allow you only to create files, but not if full access is given like here:
grafik
 
Since the Memories UI already has a button to delete selected files (when logged in), can you simply enable it if the share type includes the delete permission?
grafik

Similarly, in the viewer there is a delete button, that's currently not working:
grafik

Furthermore, there are also buttons that allow you to edit and rotate images, which are also leading to error messages as you can see in the screenshot above.
Of course, it would be great if those could be enabled, when the share type has the "edit" permission, but if you think that this is out of scope, I would wish that those are disabled at least.

Thank you for the great work so far, this will be a great addition to the app. I really hope that @pulsejet will merge this eventually.

gouverneurp:
Thank you for testing again, @major-mayer!
I have tried to incorporate the requested changes. The latest commits include the following changes:
- When you click on an image and are in “single image view,” the deletion button is only displayed if you have the rights to delete it. The delete button now works as well.
- The same applies to editing images.
- If you select multiple images, you can delete the selected items (if you have permission to do so).
- Three buttons are now displayed side by side at the top of the selection bar and in the header, instead of just one button. Additional buttons are then hidden in the menu (three horizontal dots).

Unfortunately, the latest commits make the PR more complicated and the list of changes longer. I hope that it is still acceptable and would now like to wait for feedback from @pulsejet.

major-mayer:
Thanks for the quick update.
Everything is working as expected, except for two things:
- When the share is created with editing permissions, the edit metadata button appears in the context menu (three dots) but doesn't work ("User is not logged in" appears on save)
- Same problem with the rotate/ mirror tool 

> Three buttons are now displayed side by side at the top of the selection bar and in the header, instead of just one button. Additional buttons are then hidden in the menu (three horizontal dots).

I like this change too, hopefully it will be accepted.


We are almost there :)

gouverneurp:
Wow, thank you for the intensive testing! It should work now:
example

major-mayer:
Looking good, thanks for the fix :)
I think now it's @pulsejet s turn...

major-mayer:
I hope this can be included in the next release. 
Otherwise, I have to consider creating a fork of this repo because I really want to start using this important feature and from my testing it works great.
Of course, it would be great if you, @pulsejet, could take a look at the changes first, to spot any implementation issues or even vulnerabilities.

pulsejet:
Very nicely done, thank you!

retnag:
Sry for necrobumping. Just wanted to say **thank you** all, I waited for this feature for years! Can't wait for it to be released.

major-mayer:
I don't think one should have to apologise for saying thank you :)
Also thanks to @pulsejet for bringing this over the finish line. I hope I wasn't too impatient

retnag:
uhm... when is this going to get released btw? I was wondering, since it seems to be done and merged. 
In my opinion this feature alone is already worth a release, is there still an issue with it?

major-mayer:
This is finally release and working as expected :partying_face:

Issue #1123, opened 2024-04-05
Sorting the albums by date of the photos


Markoise:
**Describe the solution you'd like**
The albums created should be sorted according to the date of the photos in them. Ideally, you should be able to enter a date for the album.

_Example:_
Hiking Tour with many photos on 01.05.2021.
Today (on 05.04.2024) I create an album with the photos from 01.05.2021. However, the album is sorted by creation date, although the period of the photos is in the past.

Is there any way to incorporate this?
Otherwise you would have to enter the date in the album title, but then everything becomes very confusing.

luckkmaxx:
Recently was added "Last updated" sorting, but this takes efect for the Album, not the content.

Ideally, when you add new taken photos to some albums, that albums should be reordered moving at top/bottom.

Markoise:
Well, none of this leads to a result.
I think it should be like Google Photos.
The album is sorted chronologically according to the photos in it.

That would be perfect.

AlexandruNegura:
Agree, I was just searching for this functionality since I am moving away from Google Photos. A new entry in that dropdown should be added to sort albums by the date of the last photo. I am manually creating the albums from Photos to keep location metadata and I just figured it out that this sorting is not possible and I have to re-edit each album to sort them correctly. I have ~180 albums so this will be fun :))

luckkmaxx:
Moving up!

PR #1525, opened 2025-09-01
Feature - Allow sorting of albums by date of included photos.

### LINKED ISSUE #1087: Sort albums view by its content date, not only album creation date
That is all. We can currently sort by date, but this is in fact the album creation date, not the content.

m-reiter:
Added the options to sort albums by date of their newest or oldest photo.

If this actually works, it should fix #1087 and #1123 and be a partial solution for #558.

This is my first PR, so apologies if I did anything wrong here - I'll be happy to learn and correct.


m-reiter:
@pulsejet I don't want to rush but do you already have an idea whether you'll find time to look at this in the foreseeable future? If there's anything you don't like about it, it would be good to know and maybe try fixing while my memory is still fresh. Thanks!

pulsejet:
I'm a bit concerned about the performance impact of this. The query needs to be studied carefully for effects on large albums (folks are running this with millions)

m-reiter:
Thanks for looking! If I can do anything to help, let me know.

fherreror:
Why not add a Date field to album description, manually filled by the user. In this way, it would not have to make a query for each album every time. It would sort all albums by their manually filled description date. 
Just an idea. I think this feature would be great and I think this could be an intermediate solution

PR #1546, opened 2025-10-01
Update hw-transcoding.md with TrueNAS guide

rstanuwijaya:
This PR is to update the docs to include guide to setup external transcoder on TrueNAS Scale system.


pulsejet (approved):
Thanks!

PR #1552, opened 2025-10-11
Add a download button for public shares

gouverneurp:
This PR adds a download button to public folder shares:

1.  Download button in a public share with no user signed in: image
2. After pressing the button, you can select the folder to save the images to: image
3. Images are downloaded to the directory: image


gouverneurp:
@major-mayer, are you interested in testing this as well?
Thank you for your help!

major-mayer:
Of course :) Just tested it and it works well. 
Thanks for your work, another great addition that makes Memories an even better Google Photos alternative.

PR #1580, opened 2025-12-04
Add parallel processing support for indexing files

hahagu:


pulsejet:
Not handling this complexity internally was intentional. The admin panel already describes a way to index in parallel.

```
Run index in parallel with 4 threads:
bash -c 'for i in {1..4}; do (occ memories:index &); done'
```

Issue #1581, opened 2025-12-07
Full screen select


jacksbox:
I would like to o have the possibility to select pictures in full screen mode. Via button and a keyboard shortcut (important to create a easy usage flow). Ideally the selections can be started / continued in the gallery view.

Why I want this: I often have to select pictures I want to group together in an album, or select pictures for printing, deleting etc. but especially for picture sets where there are a lot of similar pictures (very common today I would say) I have to open fullscreen, search for the best one, go back into gallery and select / move / delete / etc. ... And then th next set of similar pictures. It would be way smoother if I could open the full screen view, flip through all pictures and just hit a key on the ones I want to select. And in the end I only need to operate on the selection just once.

PR #1584, opened 2025-12-13
Proposal: Allow Selection in Viewer

jacksbox:
I create a proposal how selecting in the full screen viewer window could achieved (see my issue https://github.com/pulsejet/memories/issues/1581). It retains the selection when switching between viewer and timeline view.

The PR would still need some smoothing down, there are to lint erros and there are two issues I still need to solve:

* selecting an image in the viewer with no previous selected image flags the image correctly as selected in the viewer, and if I log the state of the `selectionManager` the image seems to be there. But when leaving the viewer, the image is not selected anymore. This does not happen if there is a image selected before going in the viewer mode.
* When only one image is selected, and it is deselceted in the viewer, the viewer window closes.

I have not much experience developing in vue, so I'm really happy about early feedback.


jacksbox:
@pulsejet Could you give me some feedback if this is a desired feature / general feedback to this PR? I would love to continue working on this, and/or on other memories features.

Issue #985, opened 2024-01-04
Canon Poor Exif Handling, Images Incorrectly Sorted


Civlac:
**Describe the bug**
- Sub-second time data is not accounted for when sorting Canon images, leading to some burst images being completely shuffled.
- The shutter speed is showing the "Shutter Speed Value", which can result in unfamiliar values. Should use "Exposure Time".

**To Reproduce**
- Image ordering
	- Worst case: File transfer images from local Windows 11 computer to Ubuntu Nextcloud instance via Samba. (Nextcloud configured to have External Local storage). Images can be several files out of place. (9480.jpg beside 9486.jpg)
	- Lesser extent: Upload images through Nextcloud Files web interface. This somehow orders most images better, but can swap neighbouring files' order. (9480.jpg after 9481.jpg)

**Platform:**
- OS: Windows
- Browser: Chrome
- Memories Version: 6.1.5
- Nextcloud Version: 27.1.5
- PHP Version: 8.2
- ExifTool Version Number: 12.70

**Additional context**
- For some reason, uploading the images through the online Nextcloud Files interface orders the images _better_ than moving them to Nextcloud Ext storage via Samba, despite the metadata appearing nearly identical, aside from "File Access Date/Time" and "File Inode Change Date/Time".
- All Canon models I've got my hands on have "Sub Sec Time" / "Sub Sec Original" / "Sub Sec Digitized", or "Create Date" / "Date/Time Original" / "Modify Date" that include subsecond data to the hundredths.
- [Additional detail for "Shutter Speed Value" vs "Exposure Time"](https://www.dpreview.com/forums/post/27645854)

**Feature Request**
- Within the Edit Metadata window, allow the user to modify the ms/subsecond value of timestamps to fix image sorting inconsistencies.

Civlac:
Additionally, since not all cameras appear to have sub-second time data in the exif, has it been considered to make use of the filename for these instances?

pulsejet:
Using the filename sounds like a good solution for now (though it wouldn't close this issue)

Civlac:
Sounds good, and whether using Canon's Sub Sec metadata handling gets added in the future or not, other cameras, like an older Sony I looked at, doesn't have any Sub Second time data anywhere.
Using the filename as a fallback when multiple files have an equal date and time sounds like a good fix for all images, beyond the scope of Canon.

pulsejet:
[...]

PR #1588, opened 2025-12-17
Improved EXIF Parsing

### LINKED ISSUE #815: Videos have incorrect time on timeline
**Describe the bug**
Video files show up in the wrong order in the timeline, only during summertime. The time at the info tab doesn't match up with the real time it was taken at and doesn't match up with ordering in the timeline either. Depening on the camera it's either sorted correctly but displays the wrong time in the info tab or displays the wrong time  at -1h but gets sorted as +1h. Files from before March during wintertime get ordered correctly.

**To Reproduce**
Check file order of a video and a corresponding photo from the same time. The photo has the correct timestamp, the video doesn't and gets ordered incorrectly. (sample files sent via Discord)

**Platform:**

- OS: Windows
- Browser: Firefox, Edge
- Memories Version: 5.4.1
- Nextcloud Version: 27.0.2.  AIO Version 27.1.1 beta
- PHP Version: 8.1.23

- Any errors in the Nextcloud server logs?
Probably not related but I found this:
[webdav] Fehler: Sabre\DAV\Exception: Unknown error while seeking content at <<closure>>

0. /var/www/html/apps/dav/lib/BulkUpload/MultipartRequestParser.php line 129
   OCA\DAV\BulkUpload\MultipartRequestParser->isAt("--boundary_.oOo ... n")
1. /var/www/html/apps/dav/lib/BulkUpload/BulkUploadPlugin.php line 71
   OCA\DAV\BulkUpload\MultipartRequestParser->isAtLastBoundary()
2. /var/www/html/3rdparty/sabre/event/lib/WildcardEmitterTrait.php line 89
   OCA\DAV\BulkUpload\BulkUploadPlugin->httpPost(["Sabre\\HTTP\\Request"], ["Sabre\\HTTP\\Response"])
3. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 472
   Sabre\DAV\Server->emit("method:POST", [["Sabre\\HTTP\\ ... ]])
4. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 253
   Sabre\DAV\Server->invokeMethod(["Sabre\\HTTP\\Request"], ["Sabre\\HTTP\\Response"])
5. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 321
   Sabre\DAV\Server->start()
6. /var/www/html/apps/dav/lib/Server.php line 364
   Sabre\DAV\Server->exec()
7. /var/www/html/apps/dav/appinfo/v2/remote.php line 35
   OCA\DAV\Server->exec()
8. /var/www/html/remote.php line 172
   require_once("/var/www/html/a ... p")

POST /remote.php/dav/bulk
from 83.135.74.59 by EFA at 2023-09-11T13:10:00+00:00

### LINKED ISSUE #1144: Memories Android photos lacking EXIF date/time are not sorted properly in the timeline
**Describe the bug**
Local photos in the memories Android app with no date in the EXIF data get sorted to 7:00PM Dec 31, 1969. Once they are uploaded, a duplicate will be created which is placed properly, according to the date modified/created (not sure which) info for the file. Photos which have a date/time in the EXIF data are unaffected.

**To Reproduce**
Steps to reproduce this behavior:
1. Enable local photos in the Android app
2. Add a photo to your library with no date in the EXIF data (for me, any photo saved from Signal messenger or any screenshot)
3. It will be sorted to Dec 31, 1969 at 7:00 PM
4. Upload photo (using Nextcloud Android app auto upload)
5. Duplicate photo appears in Memories Android, at the proper date/time

**Platform:**

- OS: GrapheneOS Android 14
- Browser: N/A
- Memories Android Version: 1.10
- Nextcloud Version: 28.0.4
- PHP Version: N/A

### LINKED ISSUE #1495: Timezone is not taken into account when sorting by date
### Describe the bug

In the timeline, it looks like Memories sorts images using only the time, without taking into account the timezone.

I was in a trip and took photos with my phone and with a camera. The phone was set to the correct timezone, but I forgot to update the camera's settings. It shouldn't matter, since both devices had the correct time set, it is only that one of them was set to a different timezone. Example: the phone displayed 15:00 UTC-4 and the camera 21:00 UTC+2, which is the same timestamp. However, when sorting, the timezone is not taken into account and Memories thinks 15:01 UTC-4 < 15:02 UTC+2, even if the second timestamp is 6 hours earlier.

If it matters, the phone is an iPhone 15 and the file format is HEIC, while the camera is a Canon R7 and the file format is TIFF.

### Steps To Reproduce

1. Upload 2 photos with the following timestamps:
    a. 19:00 UTC-4
    b. 19:01 UTC+2
2. Check the sorting in the timeline. Memories will think photo A was taken before photo B, but B was actually taken 6 hours earlier.

Here is a link with two photos that reproduce the issue:
https://drive.google.com/file/d/1R7TY78nIwsIPgkC8il5r3rQvLWvmfRUU/view?usp=sharing

### Platform

```markdown
- OS:Linux
- Browser: Firefox
- Memories Version: 7.6.1
- Nextcloud Version: 31.0.7
- PHP Version: 8.3.23
```

### Screenshots

_No response_

### Additional context

_No response_

### LINKED ISSUE #1569: Shared album sort order should all be based on files "Created date" instead of files "Modified date"
When I fine-tune a photo or compress a video then overwrite the files (with same filenames) on my Nextcloud instance, Memories will change the sort order of the photos & videos based on the files "Modified date": would be solved if Memories sort order are always based on files "Created date" instead of files "Modified date" (both for Timeline & Shared Albums) .

As foto & video maker when I share an album of photos & videos, I want that the order of the photos & videos always remain the same, even if I modify-optimize some photos or videos in the shared album.

There is currently no alternative solution to solve this behavior, impacting both the "Timeline" and "Shared Albums".

SilentNightx:
## Related Issues
- Fixes #1569
- Fixes #1495
- Fixes #1144
- Fixes #815
- Related to #985

## Summary
1. Now it doesn't prioritize specific fields, it instead prioritizes the oldest date found in the candidate fields. This should provide much more accurate dates and also fix the problem where if we prioritize tags commonly used for iPhone videos then Android videos have the wrong sort order and vice versa. If the oldest datetime is shared between fields it chooses the most precise one between no seconds, seconds, and subseconds.
PERFORMANCE IMPACT: Very minimal and necessary for proper sort order.

2. Uses two calls to exiftool per media item. One with -n for GPS coordinates and one without -n for anything else. I found using -n causes issues with how exiftool reports date strings with subseconds and timezones. This change allows us to support subseconds and makes it so most dates come out of exiftool with a proper timezone already attached. The only time the parser manually handles timezones now is when the oldest candidate date found lacks timezone info.
PERFORMANCE IMPACT: Very minimal. If we really want to we could reduce this to one call again if we manually parse the GPS info from the first exiftool call. Not sure which is faster.

4. Now handles more datetime EXIF fields and formats.
PERFORMANCE IMPACT: Basically none.

5. Now uses a fallback method to get timezone information from GPS coordinates if it can't be gathered from the planet database. It uses Python timezonefinder so Python 3.9+ must be installed on the server and most of the time it is by default on modern distros. This allows us to still get timezone information for Android videos (important for sorting) on SQLite instances or if the planet dataset isn't installed. The timezonefinder library and dataset is in the python directory and can be updated with pip install --upgrade timezonefinder --target /var/www/html/nextcloud/apps/memories/python. This is where the scary number of "files changed" comes from.
PERFORMANCE IMPACT: Large, but this code pretty much only runs on single-user instances, since large instances will likely be using PostgreSQL, MySQL, or MariaDB and the planet dataset. And on single-user instances without the planet dataset this code still only runs when it needs to for proper timezone gathering.

7. Now media is categorized into days based on the day they were taken in the media item timezone so they don't show on the wrong day when they are past midnight UTC.
PERFORMANCE IMPACT: Basically none.

9. Changed some UI elements to show dates based on the media item timezone.
PERFORMANCE IMPACT: Basically none.

IMPORTANT NOTE: Will need to clear/force reindex files and clear browser cache to have photos show correctly with these changes. 

If you have media that had dates previously converted to UTC from another timezone and saved back into the file EXIF dates, and didn't have its old timezone offset removed from the dedicated EXIF offset fields in the process, then reindexing it will move it off from where it should be by your timezone offset. I don't have time to look into it but somebody should double check that Memories doesn't do this when editing metadata, because that would be bad whether or not my PR is merged.

The way Date/Time metadata is edited in Memories should also eventually get some work done to bring it in line with my changes. I suggest showing the Date/Time in the media item timezone instead of UTC and adding an edit box for the timezone offset.

The way the Android app sorts local photos should also get brought in line with these changes at some point.

First contribution to a repo that isn't my own ever so go easy on me please 🙏.


executed:
@SilentNightx why I see "python" directory files added as part of this PR? Seems like they should not be here.
Thanks for this PR though!

SilentNightx:
@executed See point 4 in the summary. It's intentional. Without the python library the correct local time for some media (namely Android videos) cannot be gathered on some Nextcloud instances so the sort order for that media will be wrong.

SilentNightx:
Turns out pretty much no web or desktop clients do this other than Google Photos. They all have broken and frustrating sorting especially when it comes to Android videos. Could be a big advantage for Memories if it gets merged. @pulsejet Sorry for the ping. Just want to see if you've looked at this and #1602 at all yet. Thanks!!

pulsejet:
Hi @SilentNightx 

> Now it doesn't prioritize specific fields, it instead prioritizes the oldest date found in the candidate fields. This should provide much more accurate dates

I very much doubt this is true. The current ordering has been "learned" over a very long time from many samples. It's unclear to me what problem you're trying to solve. Is there a sample of where a wrong date is produced.

> -n causes issues with how exiftool reports date strings 
> Now handles more datetime EXIF fields and formats.
> so they don't show on the wrong day when they are past midnight UTC.
> Changed some UI elements to show dates based on the media item timezone.

Again, not clear what the issue is. Concrete samples would be helpful.

> Python 3.9+ must be installed on the server

Unfortunately we can't assume this, there are many instances where nextcloud is deployed in constrained environments. Also we should not be bundling in large python source like this.

> if the planet dataset isn't installed.

Then you're on your own basically.

SilentNightx:
@pulsejet Thank you so much for reviewing. I'll try to be more detailed about the problem. 

>The current ordering has been "learned" over a very long time from many samples. It's unclear to me what problem you're trying to solve.

Unfortunately learning from samples isn't sufficient for proper sorting since there are just too many ways that different cameras and editing software handle date and time metadata so it's impossible to cover them all properly by going off of samples and is much better to just pick the oldest datetime that exists in the fields. Some concrete examples I have of wrong sort order include photos I've edited on desktop (usually they end up at the top of the timeline instead of next to their counterpart) or photos and videos I've downloaded from someone else where the pipeline changed the metadata in the field that got prioritized by Memories but where a correct field still exists. There are also many photos I've taken on older cameras that don't get sorted well. And there are also many other examples from other users in the linked Issues.

Regarding assumption of the Python install on the server that is a simple enough issue to resolve however it causes no issues if Python isn't found. I suggest a link to install it be placed in the same place the optional planet dataset. This way single user instance SQLite users aren't left in the dark with improper Android video sorting.

Hopefully you'll consider at least some of these changes. I'm very much in need of an app with good sorting.

Thanks again!

pulsejet:
> Some concrete examples I have of wrong sort order include photos I've edited on desktop (usually they end up at the top of the timeline instead of next to their counterpart)

Please provide these examples (i.e. actual files where sorting is broken along with what it should be)

> much better to just pick the oldest datetime that exists in the fields

Without stronger evidence I find this hard to believe.

> the current Memories code uses UTC for basically everything

If the timezone cannot be inferred from the file, the server's local timezone is used instead. So I'd think either this is a bug or misconfiguration.

> This way single user instance SQLite users aren't left in the dark with improper Android video sorting.

SQLite is not recommended for Nextcloud and I especially don't recommend it for Memories for anything expect toy instances. In general, I'm not positive to taking a python dependency to solve a problem that already has a solution.

See https://docs.nextcloud.com/server/19/admin_manual/installation/system_requirements.html?highlight=sqlite#server

SilentNightx:
@pulsejet 

> Please provide these examples (i.e. actual files where sorting is broken along with what it should be)

## Here is a file, shared with me from an iPhone and downloaded to my device then uploaded to my Nextcloud instance.

Screenshot 2026-09-08 020825

The current release of Memories shows the date I downloaded it, Dec 6th, which is very far off from the date it actually belongs with. It belongs with a set from Oct 4. My PR solves this for this file and hundreds of other files in my timeline (and other users timelines too hence the upvotes and the other PR someone made to accompany this #1602)

Screenshot 2026-09-08 030654

The underlying issue in this specific case is that something in the download pipeline changes a lot of the metadata of the file to reflect the date downloaded, and Memories prioritized a changed field instead of prioritizing fields that still contain the correct date like FileModifyDate or QuickTime CreationDate.

Screenshot 2026-09-08 031230

There are too many ways that cameras, editing software, and even download servers like to manipulate metadata that you **cannot** get proper sort from sampling.

**Now that is the core of what this PR solves and none of that relies on Python.**

### Let me try to address your requests for more details on specific things better, in my last reply I was in a rush and read and responded to some things poorly.

>-n causes issues with how exiftool reports date strings

When using the -n switch with Exiftool it will not return strings with their attached timezones (or subseconds). In the release version of Memories the -n switch was used to get GPS data from Exiftool (as that's the best way to do it) but it was also used for everything else so Memories was stuck with gathering and interpreting the timezone info manually through the separate timezone metadata fields that Exiftool would usually append to datetime strings by default. By making an Exiftool call with -n and another without -n we get the best of both worlds and can let Exiftool handle most of the parsing instead of trying to have Memories do it.
https://github.com/pulsejet/memories/pull/1588/changes#diff-f0ccd1ab24a354715c83004c76196916573f3941f1363a9f416262240a87e370R92

The benefits of this include subseconds support, Exiftool being able to occasionally find timezones for files the Memories parser couldn't for some reason, and just overall reduction of complexity. Exiftool is the industry standard parser.

>Now handles more datetime EXIF fields and formats.

I've simply added more fields to the list of metadata to check for when looking for datetimes for a media item in lib/Db/ExifFields.php.

>so they don't show on the wrong day when they are past midnight UTC.

This just means I've changed the day grouping to group items by local date taken instead of UTC date, meaning items lacking timezone info no longer show grouped into the wrong day.
https://github.com/pulsejet/memories/pull/1588/changes#diff-f0ccd1ab24a354715c83004c76196916573f3941f1363a9f416262240a87e370R144

## Lastly...

> SQLite is not recommended for Nextcloud and I especially don't recommend it for Memories for anything expect toy instances. In general, I'm not positive to taking a python dependency to solve a problem that already has a solution.

Unfortunately the Nextcloud docs fail to mention that SQLite is the go to solution for single-user instances especially under WSL since other database types have had major issues under WSL. I've been using SQLite on my single-user instance for years now with no problems. Minimal yes, toy no. Running it in this fashion allows me to have absurd amounts of redundant cloud storage without paying the usual cost of that.

### I'd even go as far as to say Memories not being able to find timezones is the **only** issue with SQLite I've experienced over the years.

And again Python is only a requirement for allowing SQLite instances to find timezones and it can be made an optional download in the UI just like the planet dataset is. The rest of the benefits of this PR don't require Python.

Let me know if you want me to clarify anything else. 😀

PR #1606, opened 2026-01-25
Fixes for Android 13 and newer

### LINKED ISSUE #1605: (Bug, Android, 1.13) UI drawn under status bar + back gesture closes the app
### Describe the bug

After updating **Memories** to the latest version using the APK released in the GitHub *Releases* section, I’m experiencing two main issues related to layout and navigation:

1. The app UI is rendered **under the Android status bar**.
   - UI elements located at the top of the screen are **not clickable**.
2. Using the Android **back gesture** causes the app to **close completely** instead of navigating back within the app.

### Steps To Reproduce

1. Download and install the latest APK from the GitHub *Releases* section.
2. Open the app.
3. Notice that the UI is drawn under the status bar.
4. Try interacting with elements at the top of the screen (they don’t respond).
5. Use the Android back gesture → the app closes instead of navigating back.

### Platform

```markdown
- OS: Android 16 (OnePlus)
- Browser: - 
- Memories Version: 1.13
- Nextcloud Version: 32.0.4
- PHP Version: 8.3.27
```

### Screenshots

Image

arnowelzel:
- when the device runs Android 16 or newer, the insets for the status bar and navigation bar are applied to the main view.
- when the device runs Android 13 or newer, the "back" gesture is handled by an custom dispatcher.
- minimum SDK version is updated from 27 to 30 since the web view does not work in older versions.

This will fix #1605

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -40,10 +40,6 @@ dependencies {
     def media_version = "1.9.0"
     def room_version = "2.8.4"
 
-    implementation 'androidx.activity:activity-ktx:1.12.2'
-    implementation 'androidx.activity:activity-compose:1.12.2'
-    implementation 'androidx.compose.foundation:foundation-layout:1.10.1'
-    implementation 'androidx.compose.ui:ui:1.10.1'
     implementation 'androidx.core:core-ktx:1.17.0'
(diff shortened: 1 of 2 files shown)


pulsejet:
Thank you!

arnowelzel:
There are a few things left which should be fixed before publishing a new version. I will provide this as a second PR.

PR #1619, opened 2026-02-10
Fix public uploads for nextcloud subdirectories

gouverneurp:
This PR fixes uploads to public folders (for unregistered users) in Nextcloud instances that are deployed in a subdirectory.

For example, when a Nextcloud instance is hosted at `www.example-cloud.com/nextcloud` instead of directly at `www.example-cloud.com`, the previous implementation did not work correctly.


gouverneurp:
@major-mayer, would you be interested to test this again?
Your help was much appreciated in the past!

major-mayer:
Hi @gouverneurp ,
I am glad to hear that 😊
I could give it a try, but I guess I need a Nextcloud environment that is hosted in a subfolder, right?
My dev environment is usually just accessible via http://localhost
Probably will take a while for me

gouverneurp:
Thank you for your help!
It would be fine for me if you tested it without a subdirectory, so you can also confirm that this isn’t breaking anything.
But in the end, @pulsejet needs to decide.

major-mayer:
All right, I just tested this branch without a subdirectory hosting and the upload for external shares is still working flawlessly.
I have no objections against this PR :)

mimedo:
Does this PR fixes #1634 as well?

poggenpower:
@gouverneurp thank you for this fix. I tapped into the same issue. 
I have created a patch from this PR and applied it to version 8.0.1 of memories and it fixed the issues. 
I really would appreciate if this get merged and released.

Issue #1397, opened 2025-02-21
Choose how many days to show around current date in On This Day


Samskeli:
**Is your feature request related to a problem? Please describe.**
I have a daily habit of looking at On This Day. I have almost 40,000 photos and videos in my nextcloud, so the average daily amount of photos would be ~100, which is manageable. But because it also shows photos for 3 days before and after the current date, the average amount of photos shown is ~700/day for me, so it makes finding the "new" photos that I didn't see yesterday quite cumbersome.

**Describe the solution you'd like**
There should be an option in the Memories app to change the amount of days shown around the current date when looking at On This Day. Right now [the value](https://github.com/pulsejet/memories/blob/4644cda3e2abb864e9ebfa5b14a245b12f21a787/src/services/dav/onthisday.ts#L19) is 3 it can't be changed in the app.

**Describe alternatives you've considered**
- One alternative is that I can look at On This Day once a week on the same weekday so it always shows only photos that I haven't seen and I also don't miss any.
- The alternative that I'm currently doing is that I just change the 3s on [this line](https://github.com/pulsejet/memories/blob/4644cda3e2abb864e9ebfa5b14a245b12f21a787/src/services/dav/onthisday.ts#L19) to 0s so it only shows the photos for one date at a time. Then I build the app for myself. And look daily at On This Day.

**Additional context**
\-

hayvan96:
I second this. I love Memories, but I'm flooded with "on this Day" pics from "this week" rather than the exact day. Being able to tune the tolerance between the exact day seems practical.

PR #1621, opened 2026-02-13
Add new parameters to settings to configure 'on this day'

gouverneurp:
Adds configurable user settings for the "On This Day" feature, allowing users to customize the day range (how many days around each anniversary to include) and the number of photos shown per year. Settings can be found in the user settings menu under "On This Day".

image

Implements feature request: https://github.com/pulsejet/memories/issues/1397


gouverneurp:
@major-mayer, are you still interested in testing these new features?
Please let me know if I’m bothering you or shouldn’t tag you.

major-mayer:
Hi @gouverneurp,
yes of course you can tag me. I am always excited about new features coming to the already great Memories app.
Great that you keep working on that topic :)
I might not always reply superfast (as this time unfortunately) but eventually I will (hopefully) test it.

I tested the two new settings fields and the first one "Day range" is working as expected.
Surprisingly even leap years seem to be considered which is a nice detail.
Today (02.03.2026) It was showing me images of 26.02.2022 (regular year), but no images from 26.02.2024 (only until 27.02.2024) when I set a range of 4 days.

However, the field  Photos per year doesn't seem to have any effect in my case.
grafik
No matter what number i put here, it is always showing me over 1 image from that year.
I think this needs to be addressed first.

gouverneurp:
Hi @major-mayer,

Thank you for testing it!
Just to be sure:
- With the following settings: Screenshot from 2026-03-06 20-13-04
- You should only see 1 image per year in the timeline window. For example here when pressing on 1 year ago: Screenshot from 2026-03-06 20-13-20
- In the "On this day" section, it should be fine to display all images for the day there. In my case, its 3: Screenshot from 2026-03-06 20-12-42

What do you think?
If you still have problems, you could completly reload the page (with clearing the cache) by hitting ctrl+F5.

Thank you for your help!

major-mayer:
Hi @gouverneurp,
I think in that case everything is working as expected and I just misunderstood what this setting is supposed to do.
I thought that it's related to the "On this day" section and used to adjust how many photos are shown there.
If it should target the pictures in the top bar of the timeline, I would adjust the description to something like:
> Maximum number of photos to show per year in the top section of the timeline. Does not affect the "On this day" section (?)

Or something similar...
Maybe it would also make sense to move this setting to a different category, but since there is currently no dedicated section for "Timeline" I wouldn't know where.

One minor thing that I also noticed, is that the margins below the settings headlines vary significantly.
Compare "General" to the text input below, and "On this day" to the day range number input (the other margins are somewhere in between I would say).
It would be nice if those margins are consistent across all headlines.

PR #1666, opened 2026-05-19
Map - fix Access blocked 403r

starypatyk:
Should fix #1637

Configure the leaflet library to send the `referer` header when requesting map tiles from OSM servers.

As described in:
https://wiki.openstreetmap.org/wiki/Blocked_tiles
https://leafletjs.com/reference.html#tilelayer-referrerpolicy

Verified in my local dev environment.

### DIFF
--- a/src/components/top-matter/MapSplitMatter.vue
+++ b/src/components/top-matter/MapSplitMatter.vue
@@ -89,7 +89,7 @@ export default defineComponent({
       maxBoundsViscosity: 0.9,
     },
     tileLayerOptions: {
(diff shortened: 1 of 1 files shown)


Croydon:
@pulsejet If you have a minute eventually, the broken map feature is probably the biggest breakage in Memories right now 🙂

britzke:
Why not using "strict-origin" instead of "origin"? are we expecting an [HTTP connection downgrade](https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Referrer-Policy )?

PR #1680, opened 2026-06-11
Enable addToAlbum from album view

### LINKED ISSUE #1597: Allow adding selected photos from one album to another album

When browsing photos inside an album in Memories, there is currently no way to select a subset of photos and add them directly to another existing album.

This makes album organization cumbersome: to reuse photos in another album, I must leave the current album view and manually reselect the same photos from the timeline or folder view.  
Example: I want to select 20 photos from Album A and add them to Album B without leaving Album A.

---

Add an option within the album view that allows:
- Multi-selection of photos inside an album with a contextual action such as **“Add selected to album”**
- A dialog to choose one or more destination albums

The selected photos should be added to the target album(s) without being removed from the original album.

---

The current workaround is to go back to the timeline or folder view and manually reselect the photos to add them to another album.  
This breaks the album-based workflow and is inefficient for large collections.

---

**Additional context**

This feature would significantly improve photo organization workflows, especially for users managing large libraries with overlapping thematic albums.

It complements existing album features by enabling album-to-album reuse while preserving the original album structure.

tuxArg:
This enables adding to other albums even inside album view.

Fixes #1597

### DIFF
--- a/src/components/SelectionManager.vue
+++ b/src/components/SelectionManager.vue
@@ -271,7 +271,7 @@ export default defineComponent({
         name: t('memories', 'Add to album'),
         icon: AlbumsIcon,
         callback: this.addToAlbum.bind(this),
-        if: (self: any) => self.config.albums_enabled && !self.routeIsAlbums,
+        if: (self: any) => self.config.albums_enabled,
       },
       {
         id: 'face-move',


pulsejet:
This creates a broken experience. If you try to share a file you don't own it silently fails.

tuxArg:
@pulsejet Ok, thanks for your review. I've been using this for a while but never tested it on edge cases. I think we should add this.canShare on each file, but I'm not really familiar with the code structure here.

PR #1703, opened 2026-07-18
Support Huawei Live Photos

tablebird:
Add detection for Huawei "Live" motion photos in LivePhoto::... by inspecting JPEG trailers.


lib/Db/LivePhoto.php
```diff
@@ -131,6 +131,57 @@ public function getLivePhotoId(File $file, array $exif): string
...
+                            // Use regular dynamic capture: the number after the colon (anchor) and the negative offset after LIVE_
+                            // Matching form: 0:1477              LIVE_18666740
+                            if (preg_match('/(\d+):(\d+)\s+LIVE_(\d+)/', $trailerChunk, $matches, PREG_OFFSET_CAPTURE)) {
+
+                                $negativeOffset = (int)$matches[3][0];
+
+                                if ($negativeOffset > 0) {
+
+                                    // Calculate the absolute position of the beginning of the video according to the test formula.
+                                    $videoOffset = $fileSize - 40 - $negativeOffset + 1;
```
pulsejet, line 159: Your calculation was slightly off here :)

Here's what I ended with
```php
        // Huawei Motion Picture
        if ('image/jpeg' === ($exif['MIMEType'] ?? null) && $size > 40) {
            // LIVE_%d is the negative offset from the beggining of the
            // metadata trailer to the beginning of the video part.
            // <image> <video> <metadata: 40 bytes>
            // |0:1477              LIVE_18666740       |
            // |540:540             LIVE_4276837        |
            $trailer = file_get_contents($path, false, null, -40, null);
            if (preg_match('/LIVE_(\d+)/', $trailer ?: '', $matches)) {
                $negativeOffset = (int) $matches[1];
                $videoOffset = $size - 40 - $negativeOffset;
                if ($negativeOffset > 0 && $videoOffset > 0
                    && \in_array(file_get_contents($path, false, null, $videoOffset + 4, 4) ?: '', MP4_ATOMS, true)) {
                    return "self__traileroffset={$videoOffset}";
                }
            }
        }
```

tablebird: Your is correct, as my negligence led to the submission of incorrect code

PR #1697, opened 2026-07-14
Fix stuck loading on big video files

Leszeg:
Split Video into chunks.
Fix for issue: #1304

PR #1707, opened 2026-08-04
Fix video playback stalling until the entire file is downloaded

### LINKED ISSUE #1304: Video Playback Fails for Large Files in Memories App, Works in Files App
### Describe the bug

When attempting to play large video files (e.g., .mov) using the Memories app, the playback fails, while the same files play correctly in the Files app. Upon inspecting via Safari's web inspector, it was observed that the Memories app starts downloading the video but fails to complete the download. In contrast, the Files app downloads a portion of the video and waits for the user to press play before continuing to download and stream the file. The transcoding option is disabled as the video file is natively supported by iOS and if the size is small the file plays correctly.

### Steps To Reproduce

1. Record a large video file using iPhone 
2. Upload the video to Nextcloud
3. Playback via Memories

### Platform

```markdown
- OS: iOS 18.0
- Browser: Safari
- Memories Version: 7.4.1
- Nextcloud Version: 30.0.0 (fpm-alpine)
- PHP Version: 8.2.24
- Web server: nginx/1.26.2
```

### Screenshots

c53c938b-c5e4-432d-8608-b9fbdfed3815

VeloAddict:
Fixes #1304

### Cause
For a request with Range: bytes=0-, DownloadController::one() computes $seekStart = 0 and $seekEnd = $size - 1, so this condition is false:

```
if ($seekStart > 0 || $seekEnd < ($size - 1)) {
    $out->setHeader('HTTP/1.1 206 Partial Content');
    $out->setHeader("Content-Range: bytes {$seekStart}-{$seekEnd}/{$size}");
}
```

The response is sent as 200 OK even though a range was requested. Separately, the response carries no ETag and no Last-Modified, and Nextcloud's default Cache-Control: no-cache, no-store, must-revalidate is left in place.

A media element seeking across byte ranges needs a validator to know that a follow-up request returns the same representation. With no validator and no-store, the browser cannot resume with a second range request, so its only option is to read the single response to completion. The 200 compounds this by not advertising range support for the response at all.

Content-Encoding: none is also set, which is not a registered content coding.

### Result
The endpoint's headers now match what DAV serves. The browser buffers what it needs and drops the connection; playback starts immediately and seeking works.

Why not #1697?
#1697 caps open-ended range responses at 32 MB, which forces a 206 as a side effect of the response being partial. That works for faststart files but leaves non-faststart files unplayable — the browser receives 32 MB from the front, does not find the moov atom, and has nothing to seek toward. It also does not address the missing validator or no-store, so seeking still does not work.


VeloAddict:
@pulsejet would it be possible to include this fix in the next release?

PR #1727, opened 2026-09-21
fix(go-vod): Accelerate VA-API HDR tone mapping and preserve HLS timestamps

rafolg:
VA-API HDR transcoding can spend most of its time downloading frames for CPU tone mapping, while HLS output can independently duplicate a roughly 30 fps source to its advertised 120 fps rate. These two commits address those separate costs.

The motivating source is iPhone 15 Pro HEVC Main10, 3840×2160, HLG/BT.2020 with Dolby Vision profile 8 and −90° rotation. Its timestamps average approximately 29.98 fps, but `r_frame_rate` is `120/1`.

### VA-API/OpenCL HDR tone mapping

Before:

```text
VA-API decode → hwdownload → CPU zscale/Hable → hwupload
→ scale_vaapi → transpose_vaapi → h264_vaapi
```

After a successful interop probe:

```text
VA-API decode → scale_vaapi (P010, target resolution)
→ hwmap to OpenCL → tonemap_opencl (Hable, NV12/BT.709)
→ hwmap back to VA-API → transpose_vaapi → h264_vaapi
```

Scaling before tone mapping avoids processing pixels that a smaller rendition discards. The maximum rendition retains the source dimensions.

Before the first transcode of each HDR rendition, go-vod encodes one frame through the proposed graph into the null muxer. This exercises the actual render device, decoder, dimensions, OpenCL initialization, frame sharing, rotation and encoder. `-abort_on empty_output` prevents an empty decode from passing. A compiled-in filter or device listing alone does not establish working interop.

Failures and a 10-second timeout retain the existing CPU HDR tone mapper. The decision is cached for that stream, including concurrent requests and subsequent seeks. The existing forced software-transpose workaround retains its existing tone mapper and VA-API filter device. VA-API SDR, x264, NVENC and stream copy keep their existing graphs. The probe adds startup work and cannot guarantee against driver or decoder failures later in playback; there is no new midstream retry mechanism.

Hable is retained. No Dolby Vision decoder options are changed. Direct `tonemap_vaapi` was not selected because it failed with a processing error on the reported Intel stack.

### HLS frame timestamps

Transcoded HLS now adds the output option `-fps_mode passthrough`. FFmpeg's automatic synchronization can otherwise choose CFR for HLS/MPEG-TS and duplicate frames to the nominal rate. This is independent of tone mapping: a null-output transcode of the original source did not exhibit the same duplication.

The option is omitted for stream-copy HLS and progressive MP4. Existing `-copyts`, `+genpts`, forced keyframes/fixed GOP, `split_by_time`, seek offsets and segment numbers are retained.

### Reproduction and measurements

A generated source reproduces the HLS issue without sharing a personal video:

```sh
ffmpeg -f lavfi -i 'testsrc2=size=320x180:rate=120:duration=12' \
  -f lavfi -i 'sine=frequency=440:sample_rate=48000:duration=12' \
  -vf 'select=eq(mod(n\,8)\,0)+eq(mod(n\,8)\,3)' \
  -fps_mode vfr -c:v libx264 -threads 2 -preset ultrafast \
  -crf 28 -g 60 -c:a aac vfr-120.mp4
```

Transcode it using go-vod's generated HLS arguments. With FFmpeg 6.1.1, this 360-frame source reports `r_frame_rate=120/1` and `avg_frame_rate=10800/359`:

| HLS check | Without passthrough | With passthrough |
| --- | --- | --- |
| Full synthetic video | 1,436 frames; 1,076 duplicates | 360 frames; 0 duplicates |
| Synthetic video, seek to 6 s | 716 frames; 536 duplicates | 180 frames; 0 duplicates |
| Intel GPU/iPhone source, seek to 12 s | 720 frames; 540 duplicates | 180 frames; 0 duplicates |

The GPU seek comparison produced identical first video PTS (13.408333 after the MPEG-TS muxer offset), media sequence 4 and segment durations of 2/3/1 seconds. Its short first segment is present with both synchronization modes. Stream-copy output was unchanged in the synthetic comparison. The synthetic fixed-GOP check also preserved all 360 frames without duplication.

Earlier full-file measurements of the v9.0.1 prototype on the same Intel system reported 22,604 output frames, including 16,956 duplicates, without passthrough; with passthrough, 5,648 frames and zero duplicates. Elapsed time decreased from approximately 35.98 to 17.21 seconds. Earlier GPU tests consistently favored scaling before tone mapping; reported throughput varied by output and cache state. These are prototype measurements, not full-file benchmarks of these commits.

### Compatibility and validation

- Based implementation on current master, `9f60d223`. The relevant argument builder is unchanged from v9.0.1. Existing capability checks only locate binaries/report versions; they do not verify filters or interop.
- **HLS transcoding requires FFmpeg 5.1+.** `fps_mode` is present in [FFmpeg 5.1](https://github.com/FFmpeg/FFmpeg/blob/n5.1/fftools/ffmpeg_opt.c). The repository's oldest pinned FFmpeg build is [5.1.3 for NVENC](https://github.com/pulsejet/memories/blob/9f60d223/go-vod/build-ffmpeg-nvidia.sh); the VA-API script pins 6.0. The existing documentation has no enforced precise minimum. This PR documents 5.1+ explicitly; it does not add a legacy `-vsync` fallback for older installations.
- `p010` resolves to native-endian P010, including `p010le` on the tested machine, via [FFmpeg's pixel-format lookup](https://ffmpeg.org/doxygen/5.1/pixdesc_8c_source.html). The OpenCL filter accepts P010 input; unsupported negotiations fail the runtime probe and retain the previous graph.
- Baseline `go test ./...` passed. Final `go test -race ./...` and `CGO_ENABLED=0 go build` passed with Go 1.27.0. Go files were formatted and `git diff --check` passed.
- Regression tests cover GPU graph order and rotation, original-size output, unaffected backends, probe failure/cancellation, concurrent per-rendition caching, skipped probes, and HLS option placement/copy exclusion.
- Isolated hardware validation of these commits' probe and argument builder: Intel UHD 770 (`8086:a780`), i915, TrueNAS kernel `6.12.105-production+truenas`, `/dev/dri/renderD128`, FFmpeg `8.1.2-Jellyfin` from the saved `radialapps/go-vod`-based image. 480p, 720p, 1080p, 1440p, max and a 12-second seek each encoded 180 frames with zero duplicates/drops and BT.709 output. Successful probes took approximately 0.35–0.83 seconds in the rendition run. This validates generated commands and full browser playback session of the upstream daemon.
- The reported live installation uses Memories 9.0.1/go-vod 0.4.2. Hardware validation ran separately from that restored production container.
- `make php-lint` could not run: the checkout lacks `vendor/bin/php-cs-fixer`. No PHP files changed. No NVENC/AMD hardware or FFmpeg 5.1 runtime was available for direct validation.


docs/hw-transcoding.md
```diff
@@ -16,6 +20,20 @@ To configure VAAPI, you need to have `/dev/dri` available to the Nextcloud insta
 
 NVIDIA GPUs support hardware transcoding using NVENC.
 
+For VA-API HDR sources, go-vod tries a single-frame transcode before starting
+each rendition. When VA-API/OpenCL frame sharing works, it scales the 10-bit
```
pulsejet, line 24: This is admin documentation, pls remove the implementation details from here

go-vod/ffmpeg/args.go
```diff
@@ -298,6 +325,10 @@ func tonemapFilter(cv, scaler string, scalerArgs []string) string {
 //     keyframe past each -hls_time boundary, matching CopySegments.
 func SegmentArgs(s Spec, startID int, pattern string) []string {
 	args := BuildArgs(s)
+	if !s.Copy {
+		// HLS may otherwise duplicate VFR frames to match the probed r_frame_rate.
+		args = append(args, "-fps_mode", "passthrough")
+	}
```
pulsejet, line 331: move this into the dedicated `!s.Copy` block below

go-vod/ffmpeg/args.go
```diff
@@ -263,6 +276,20 @@ func BuildArgs(s Spec) []string {
 	return args
 }
 
+// Scale HDR surfaces before tonemapping to avoid processing discarded pixels.
+func vaapiTonemapFilter(scalerArgs []string) string {
+	args := slices.Clone(scalerArgs)
+	for i, arg := range args {
+		if arg == "format=nv12" {
+			args[i] = "format=p010"
```
pulsejet, line 284: This should be changed at source rather than patching like this

pulsejet (changes requested):

go-vod/ffmpeg/args_test.go
```diff
@@ -149,6 +151,53 @@ func TestBuildArgsNoAudio(t *testing.T) {
...
+				want += ",transpose_vaapi=1,transpose_vaapi=1"
+			}
+			require.Equal(t, want, filter)
+			require.NotContains(t, strings.Join(args, " "), "hwdownload")
+			require.NotContains(t, filter, "zscale")
+		}
+	}
+}
+
+func TestVAAPIOpenCLDoesNotChangeOtherBackends(t *testing.T) {
```
pulsejet, line 187: please drop this test

pulsejet:
@rafolg I plan to fixup this PR and merge, no need to address the comments.

pulsejet:
merged in df5bbc7cd22c9c59b584ab65f5c5d8c079d83a72

